PR #2, opened 2018-05-11
Add auth module with Hudson client which gets access token

acroz:
This PR implements a Hudson client with a single method getting an access token for SherlockML.

This PR is intended to prototype:

* Package structure (specifically, having clients in a `clients` subpackage)
* A pattern for HTTP client classes
* Test configuration, including travis integration
* Use of `requests_mock` for testing HTTP client code

Note that `requests_mock` has a pytest fixture on its master branch, but this is not yet released.


sherlockml/clients/auth.py
```diff
@@ -0,0 +1,48 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from datetime import datetime, timedelta
+import pytz
+from collections import namedtuple
```
srstevenson: Can we reorder these to have third party imports (`pytzt`) after standard library imports (`collections`)?

sherlockml/clients/auth.py
```diff
@@ -0,0 +1,48 @@
...
+
+import requests
+
+
+AccessToken = namedtuple('AccessToken', ['token', 'expires_at'])
+
+
+class HudsonClient(object):
```
srstevenson: (I discussed this in person with @acroz, but am noting it here for the benefit of others).

`Hudson` is an internal reference that won't make sense to external consumers of this library. I would prefer to name this `AuthClient` or similar, to aid discoverability and understanding.

sherlockml/clients/auth.py
```diff
@@ -0,0 +1,48 @@
...
+    """Session with the SherlockML authentication service."""
+
+    def __init__(self, url):
+        self._session = requests.Session()
+        self.url = url
+
+    def get_access_token(self, client_id, client_secret):
```
srstevenson: Do we expect the same client to be used for multiple identities? I would have expected the client ID and secret to be instance attributes passed at instantiation.

zblz: I think having both url and self.url might be confusing. Could either call this variable endpoint or change self.url to , e.g,self.host.

zblz:
The `requests_mock` fixture is available in requests_mock 1.5.0, which was released 18 hours ago (and it's the version that was pulled by Travis). We should probably use that in tests if this PR is to be a model for future tests.

sherlockml/clients/auth.py
acroz: Done.

acroz:
@zblz yes that is good timing - when I was checking this module out on Thursday night that had not been released! Have now updated it to use it.

sherlockml/clients/auth.py
acroz: @srstevenson - I live having the arguments of the method reflect the parameters of the API request, since the client method is intended to encapsulate those requests in code. Also, I would like to encourage less reliance on state in classes where possible, for testing / maintenance benefits.

sherlockml/clients/auth.py
acroz: Done - renamed to `AccessTokenClient`

acroz:
All existing comments have been addressed - any other input before we merge this?

PR #5, opened 2018-05-18
SherlockML auth implementation for requests

acroz:
This PR adds an implementation of the request auth interface for performing authenticated requests to SherlockML services. This will later be used in clients implemented in this library.

This PR deliberately does not implement any on-disk caching of the access token - this will be done in a later PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .mypy_cache/
 .pytest_cache/
 __pycache__/
+*.pyc
--- a/sherlockml/clients/auth.py
+++ b/sherlockml/clients/auth.py
@@ -67,14 +67,18 @@ class SherlockMLAuth(AuthBase):
     >>> auth = SherlockMLAuth('https://hudson.example.sherlockml.net',
                               your_client_id, your_client_secret)
 
-    then pass it as the ``auth`` argument when constructing a ``requests``
-    session:
+    then pass it as the ``auth`` argument when making a request with
+    ``requests``:
+
+    >>> import requests
+    >>> requests.get('http://service.example.sherlockml.net', auth=auth)
+
+    You can also set it as the ``auth`` attribute on a
+    :class:`requests.Session`, so that subsequent requests will be
+    authenticated automatically:
 
     >>> import requests
     >>> session = requests.Session(auth=auth)
-
-    subsequent requests made using this session will be authenticated
-    automatically.
     """
 
     def __init__(self, auth_service_url, client_id, client_secret):
--- a/sherlockml/clients/auth.py
+++ b/sherlockml/clients/auth.py
@@ -17,6 +17,7 @@ from collections import namedtuple
 from datetime import datetime, timedelta
 
 import requests
+from requests.auth import AuthBase
 import pytz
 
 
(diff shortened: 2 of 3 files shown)


sherlockml/clients/auth.py
```diff
@@ -46,3 +47,54 @@ def get_access_token(self, client_id, client_secret):
...
+    construct an instance of this class:
+
+    >>> auth = SherlockMLAuth('https://hudson.example.sherlockml.net',
+                              your_client_id, your_client_secret)
+
+    then pass it as the ``auth`` argument when constructing a ``requests``
+    session:
+
+    >>> import requests
+    >>> session = requests.Session(auth=auth)
```
janfreyberg: When I try this, I get an error:

```python
------------------------------------------------------------------
TypeError                        Traceback (most recent call last)
<ipython-input-11-7e7ea1c88d65> in <module>()
----> 1 requests.Session(auth=authenticator)

TypeError: __init__() got an unexpected keyword argument 'auth'
```

Don't you want to use this syntax from the [requests docs](http://docs.python-requests.org/en/master/user/advanced/):

```python
s = requests.Session()
s.headers.update({'x-test': 'true'})

# both 'x-test' and 'x-test2' are sent
s.get('http://httpbin.org/headers', headers={'x-test2': 'true'})
```

Given that authentication occurs via headers, this would be sufficient, no?

janfreyberg: (although note this would lose the ability to update the token)

zblz: I think the syntax for setting the auth mechanism on a session is
```
s = requests.Session()
s.auth = SherlockMLAuth(...)
```

janfreyberg: yes, this works for me

sherlockml/clients/auth.py
```diff
@@ -17,6 +17,7 @@
 from datetime import datetime, timedelta
 
 import requests
+from requests.auth import AuthBase
 import pytz
```
srstevenson: Could we make this import ordering be:
```python
import pytz
import requests
from requests.auth import AuthBase
```

sherlockml/clients/auth.py
```diff
@@ -46,3 +47,54 @@ def get_access_token(self, client_id, client_secret):
...
+
+class SherlockMLAuth(AuthBase):
+    """Requests auth implementation for accessing SherlockML services.
+
+    Parameters
+    ----------
+    auth_service_url : str
+        The URL of the SherlockML authentication service
+    client_id : str
+        The SherlockML client ID to use for authentication
```
srstevenson: I think due to the context we could drop `SherlockML` from this parameter description.

sherlockml/clients/auth.py
```diff
@@ -46,3 +47,54 @@ def get_access_token(self, client_id, client_secret):
...
+    client_secret : str
+        The client secret associated with the client ID
+
+    To perform an authenticated request against a SherlockML service, first
+    construct an instance of this class:
+
+    >>> auth = SherlockMLAuth('https://hudson.example.sherlockml.net',
+                              your_client_id, your_client_secret)
+
+    then pass it as the ``auth`` argument when constructing a ``requests``
```
srstevenson: Is it useful to use the Sphinx reference notation for this?
```
:class:`requests.Session`
```

sherlockml/clients/auth.py
acroz: Hmm, I was convinced you could pass auth as a keyword argument at construction time. Will fix this.

sherlockml/clients/auth.py
acroz: Have implemented this.

sherlockml/clients/auth.py
acroz: Done.

sherlockml/clients/auth.py
acroz: Sure, done.

sherlockml/clients/auth.py
acroz: Possibly, though I thought in this case being as unambiguous as possible might be advantageous.

I'm happy to drop it though if people feel that is better - I expect this class will not be used directly by casual users, but rather only by people developing clients in the library.

sherlockml/clients/auth.py
```diff
@@ -46,3 +47,58 @@ def get_access_token(self, client_id, client_secret):
...
+
+    >>> import requests
+    >>> requests.get('http://service.example.sherlockml.net', auth=auth)
+
+    You can also set it as the ``auth`` attribute on a
+    :class:`requests.Session`, so that subsequent requests will be
+    authenticated automatically:
+
+    >>> import requests
+    >>> session = requests.Session(auth=auth)
```
janfreyberg: this is still the wrong syntax, no?

acroz: Yes, actually fixed it now.

acroz:
@sbalian Eventually, probably yes, however for the moment I suggest we don't document things extensively until the interfaces are clear and are unlikely to change. I documented this case as it's not obvious without usage information how this class is used to perform authenticated requests otherwise.

.gitignore
```diff
@@ -3,3 +3,4 @@
 .mypy_cache/
 .pytest_cache/
 __pycache__/
+*.pyc
```
sbalian: @acroz I add this. I think you were not seeing those pyc's because you're ignoring them in your global git config?

acroz: Well I think it's just that I didn't run the code in a way that made .pyc files in the main source directory. I don't have it in my global git ignores. It's useful to include though, I'll add it.

srstevenson: I agree with @acroz on not pushing without coordination.

PR #6, opened 2018-05-18
Add build indicator

acroz:
This displays the result of the last test on master in the README and provides a visual indication that we are running automated tests on the codebase.


pbugnion:
For reference, this is what the README looks like with this PR:

screen shot 2018-05-18 at 14 31 51

PR #7, opened 2018-06-11
Configuration module

acroz:
This PR implements a module for reading configuration, including a function for resolving:

* Configuration passed in programatically (called 'overrides' in the code)
* Configuration set with environment variables
* Configuration read from file
* Configuration defaults

This would be used in client code as follows:

```python
from sherlockml import config

class SomeClient(object):

    def __init__(self, configuration_path=None, profile_name=None, domain=None,
                 protocol=None, client_id=None, client_secret=None):

        profile = config.resolve_profile(
            configuration_path_override=configuration_path,
            profile_name_override=profile_name,
            domain_override=domain,
            protocol_override=protocol,
            client_id_override=client_id,
            client_secret_override=client_secret
        )
        # Use or store the profile
```

This therefore enables a client API similar to boto, where by default, configuration has defaults, and configuration and credentials are read from the environment and configuration files, but can be overridden in code.

The configuration files read here are backwards compatible with `sml`, but to avoid ambiguity, the 'SherlockML env' concept is dropped and replaced with a 'SherlockML profile'. This is more similar to boto, and importantly avoids the current double use of `SHERLOCKML_ENV` for selecting both different domains and credential sets. An example file would be:

```ini
[default]
domain = sherlockml.com
protocol = https
client_id = my-client-id
client_secret = my-client-secret
```

### DIFF
--- a/sherlockml/config.py
+++ b/sherlockml/config.py
@@ -5,12 +5,12 @@ from functools import wraps
 from six.moves.configparser import ConfigParser, NoSectionError, NoOptionError
 
 
-Config = namedtuple('Config', ['default_deployment', 'deployments'])
-Deployment = namedtuple(
-    'Deployment',
+Profile = namedtuple(
+    'Profile',
     ['domain', 'protocol', 'client_id', 'client_secret']
 )
 
+DEFAULT_PROFILE = 'default'
 DEFAULT_DOMAIN = 'sherlockml.com'
 DEFAULT_PROTOCOL = 'https'
 
@@ -27,22 +27,17 @@ def load(path):
         except (NoSectionError, NoOptionError):
             return fallback
 
-    default_deployment = _get('default', 'deployment')
-    deployments = {}
+    profiles = {}
 
     for section in parser.sections():
-
-        if section.lower() == 'default':
-            continue
-
-        deployments[section] = Deployment(
+        profiles[section] = Profile(
             domain=_get(section, 'domain', DEFAULT_DOMAIN),
             protocol=_get(section, 'protocol', DEFAULT_PROTOCOL),
             client_id=_get(section, 'client_id'),
             client_secret=_get(section, 'client_secret')
         )
 
-    return Config(default_deployment, deployments)
+    return profiles
 
 
 def env_override(environment_variable):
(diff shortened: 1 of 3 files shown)


tests/test_config.py
```diff
@@ -0,0 +1,203 @@
...
+                 return_value=CREDENTIALS_ONLY_PROFILE)
+    profile = config.resolve_profile()
+    assert profile == config.Profile(
+        domain=config.DEFAULT_DOMAIN,
+        protocol=config.DEFAULT_PROTOCOL,
+        client_id='test-client-id',
+        client_secret='test-client-secret'
+    )
+
+
```
zblz: We might want to check on these two last tests that the `CredentialsError` contains the credential type that was missing.

sherlockml/config.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class CredentialsError(RuntimeError):
+    pass
+
+
+def _missing_credentials(type_):
+    raise CredentialsError('No {} found'.format(type_))
+
+
```
zblz: I'd probably drop the `_override` suffix from these arguments. It seems easier for the code that calls `resolve_profile` to use the `domain` argument when setting the domain. The fact that it overrides all other sources does not need to be made explicit in the name.

tests/test_config.py
```diff
@@ -0,0 +1,203 @@
...
+    assert config.load('does-not-exist') == {}
+
+
+@pytest.mark.parametrize('profile_name, expected_profile', [
+    ('default', DEFAULT_PROFILE),
+    ('missing profile', EMPTY_PROFILE)
+])
+def test_load_profile(mocker, profile_name, expected_profile):
+    mocker.patch('sherlockml.config.load', return_value=SAMPLE_CONFIG)
+    path = mocker.Mock()
```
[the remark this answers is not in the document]
acroz: Changed to a string.

tests/test_config.py
acroz: Done.

sherlockml/config.py
acroz: Done.

acroz:
@zblz Yes what you said about the `__init__` method of clients that use this is true, however I imagine we may have a common `__init__` in a base class. We should discuss this when implementing the actual code, though.

PR #12, opened 2018-07-01
Add project client

acroz:


sherlockml/clients/project.py
```diff
@@ -0,0 +1,28 @@
+from collections import namedtuple
```
zblz: This file is missing the copyright blurb.

PR #13, opened 2018-07-01
Add server client

acroz:
This PR implements get and list methods.

### DIFF
--- a/sherlockml/clients/base.py
+++ b/sherlockml/clients/base.py
@@ -108,3 +108,11 @@ class BaseClient(object):
         response = self._get_raw(endpoint, **kwargs)
         _check_status(response)
         return _deserialise_response(schema, response)
+
+    def _post_raw(self, endpoint, *args, **kwargs):
+        return self._request('POST', endpoint, *args, **kwargs)
+
+    def _post(self, endpoint, schema, **kwargs):
+        response = self._post_raw(endpoint, **kwargs)
+        _check_status(response)
+        return _deserialise_response(schema, response)
--- a/sherlockml/clients/server.py
+++ b/sherlockml/clients/server.py
@@ -80,9 +80,41 @@ class ServerSchema(BaseSchema):
         )
 
 
+class ServerIdSchema(BaseSchema):
+
+    instanceId = fields.UUID(required=True)
+
+    @post_load
+    def make_server_id(self, data):
+        return data['instanceId']
+
+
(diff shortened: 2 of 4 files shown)


sherlockml/clients/base.py
```diff
@@ -108,3 +108,11 @@ def _get(self, endpoint, schema, **kwargs):
         response = self._get_raw(endpoint, **kwargs)
         _check_status(response)
         return _deserialise_response(schema, response)
+
+    def _post_raw(self, endpoint, *args, **kwargs):
```
sbalian: Since this function has a one-line body and is only being used once below in `_post`, can't we just incorporate its body into `_post`?

sherlockml/clients/server.py
```diff
@@ -0,0 +1,126 @@
...
+    projectId = fields.UUID(required=True)
+    ownerId = fields.UUID(required=True)
+    name = fields.Str(required=True)
+    instanceType = fields.Str(required=True)
+    milliCpus = fields.Int(required=True)
+    memoryMb = fields.Int(required=True)
+    createdAt = fields.DateTime(required=True)
+    status = EnumField(ServerStatus, by_value=True, required=True)
+    services = fields.Nested(ServiceSchema, many=True, required=True)
+
```
sbalian: How about `destroyed_at`?

sherlockml/clients/server.py
```diff
@@ -0,0 +1,126 @@
...
+        payload = {
+            'instanceType': server_type,
+            'milliCpus': milli_cpus,
+            'memoryMb': memory_mb
+        }
+        if name:
+            payload['name'] = name
+        if image_version:
+            payload['typeVersion'] = image_version
+        if initial_environment_ids:
```
sbalian: For interactions with other services, should we use client instances instead of client properties directly? In other words, replace `initial_environments_ids` with a container `initial_environments` that holds instances of the environment client.

sherlockml/clients/server.py
srstevenson: Galleon doesn't return that in the payload.

sbalian: Thanks @srstevenson , I was thinking that would be useful when calculating costs.

sherlockml/clients/base.py
acroz: We could, although the code implemented here is consistent with how the `get` methods are currently implemented.

PR #28, opened 2018-08-23
Format code with black

acroz:
This PR formats the code with black and adds CI checks to check black formatting.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,37 +1,39 @@
 language: python
 cache: pip
 install:
-  - pip install flake8 black
+  - pip install flake8
 jobs:
   include:
+    - name: Check black formatting
+      stage: test
+      install:
+        - pip install black
+      script:
+        - black --check .
+      python: "3.6"
     - stage: test
       script:
         - flake8
-        - black --check .
         - python setup.py test
       python: "2.7"
     - stage: test
       script:
         - flake8
-        - black --check .
         - python setup.py test
       python: "3.4"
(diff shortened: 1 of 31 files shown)


.travis.yml
```diff
@@ -1,32 +1,37 @@
 language: python
 cache: pip
 install:
-  - pip install flake8
+  - pip install flake8 black
```
srstevenson: I don't think installing Black in the install step of each job will work, as Black needs Python >=3.6 to run and we have jobs for 2.7, 3.4, and 3.5.

acroz: What do you think the right solution is? I was thinking of just having a completely separate job that does the flake8 and black checks that runs independently of the actual code checks themselves. Is there value in running flake8 on all versions?

srstevenson: A separate job that just checks formatting with Black sounds good.

I'm in favour of still running flake8 with each Python version though: the [flake8 documentation](http://flake8.pycqa.org/en/latest/) says:
> It is very important to install Flake8 on the correct version of Python for your needs. If you want Flake8 to properly parse new language features in Python 3.5 (for example), you need it to be installed on 3.5 for Flake8 to understand those features. In many ways, Flake8 is tied to the version of Python on which it runs.

acroz: Fair enough - can we go a per-version 'install' step for black so it appears as 'error' if the install fails?

srstevenson: I think so.

PR #29, opened 2018-10-03
Report client

janfreyberg:
This implements a report client that can retrieve information about reports and create new reports

### DIFF
--- a/sherlockml/clients/report.py
+++ b/sherlockml/clients/report.py
@@ -14,7 +14,7 @@
 
 from collections import namedtuple
 
-from marshmallow import Schema, fields, post_load, pre_load
+from marshmallow import Schema, fields, post_load, EXCLUDE
 
 from sherlockml.clients.base import BaseClient
 
@@ -22,7 +22,7 @@ Report = namedtuple(
     "Report", ["created_at", "name", "id", "description", "active_version"]
 )
 
-VersionedReport = namedtuple(
+ReportWithVersions = namedtuple(
     "VersionedReport",
     [
         "created_at",
@@ -44,17 +44,11 @@ ReportVersion = namedtuple(
         "report_bucket",
         "notebook_path",
         "id",
-        "report_id",
     ],
 )
 
 
 class ReportVersionSchema(Schema):
-    def __init__(self, report_id=None, *args, **kwargs):
-        super(ReportVersionSchema, self).__init__(*args, **kwargs)
-        if report_id is not None:
-            self.context["report_id"] = report_id
-
     id = fields.UUID(data_key="version_id", required=True)
     created_at = fields.DateTime(required=True)
     author_id = fields.UUID(required=True)
@@ -63,12 +57,6 @@ class ReportVersionSchema(Schema):
     report_bucket = fields.String(required=True)
     notebook_path = fields.String(required=True)
     report_id = fields.UUID()
-
-    @pre_load
-    def set_report_id(self, data):
-        if data.get("report_id") is None and "report_id" in self.context:
-            data["report_id"] = self.context["report_id"]
-        return data
 
     @post_load
     def make_report_version(self, data):
@@ -82,19 +70,12 @@ class ReportSchema(Schema):
     description = fields.String(required=True)
     active_version = fields.Nested(ReportVersionSchema, required=True)
 
-    @pre_load
-    def add_report_id_to_context(self, data):
-        if "report_id" in data:
-            self.context["report_id"] = data["report_id"]
-        return data
-
     @post_load
     def make_active_report(self, data):
         return Report(**data)
 
 
-class VersionedReportSchema(Schema):
-
+class ReportWithVersionsSchema(Schema):
     created_at = fields.DateTime(required=True)
     name = fields.String(required=True, data_key="report_name")
     id = fields.UUID(required=True, data_key="report_id")
@@ -102,22 +83,9 @@ class VersionedReportSchema(Schema):
     active_version_id = fields.UUID(required=True)
     versions = fields.Nested(ReportVersionSchema, required=True, many=True)
 
-    @pre_load
-    def add_report_id_to_context(self, data):
-        if "report_id" in data:
-            self.context["report_id"] = data["report_id"]
-        return data
-
     @post_load
     def make_versioned_report(self, data):
-        return VersionedReport(**data)
-
-
-def _relative_project_path(path):
-    if path.startswith("/project/"):
-        return path[8:]
-    else:
-        return path
+        return ReportWithVersions(**data)
 
 
 class ReportClient(BaseClient):
@@ -128,9 +96,9 @@ class ReportClient(BaseClient):
         endpoint = "/project/{}".format(project_id)
         return self._get(endpoint, ReportSchema(many=True))
 
-    def list_report_versions(self, report_id):
+    def get_with_versions(self, report_id):
         endpoint = "/report/{}/versions".format(report_id)
-        return self._get(endpoint, VersionedReportSchema())
+        return self._get(endpoint, ReportWithVersionsSchema())
 
     def get(self, report_id):
         endpoint = "/report/{}/active".format(report_id)
@@ -138,20 +106,19 @@ class ReportClient(BaseClient):
 
     def create(
         self,
+        project_id,
         name,
         notebook_path,
-        project_id,
         author_id,
         description=None,
         show_code=False,
     ):
-        notebook_path = _relative_project_path(notebook_path)
 
         payload = {
             "report_name": name,
-            "author_id": author_id,
-            "notebook_path": notebook_path,
-            "description": description if description is not None else "",
+            "author_id": str(author_id),
+            "notebook_path": str(notebook_path),
+            "description": description,
             "show_input_cells": show_code,
         }
 
@@ -162,13 +129,13 @@ class ReportClient(BaseClient):
         self, report_id, notebook_path, author_id, show_code=False
     ):
 
-        notebook_path = _relative_project_path(notebook_path)
-
         payload = {
-            "notebook_path": notebook_path,
-            "author_id": author_id,
+            "notebook_path": str(notebook_path),
+            "author_id": str(author_id),
             "show_input_cells": show_code,
         }
 
         endpoint = "/report/{report_id}/version".format(report_id=report_id)
-        return self._post(endpoint, ReportVersionSchema(), json=payload)
+        return self._post(
+            endpoint, ReportVersionSchema(unknown=EXCLUDE), json=payload
+        )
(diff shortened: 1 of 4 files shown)


sherlockml/clients/report.py
```diff
@@ -0,0 +1,174 @@
...
+        "report_id",
+    ],
+)
+
+
+class ReportVersionSchema(Schema):
+    def __init__(self, report_id=None, *args, **kwargs):
+        super(ReportVersionSchema, self).__init__(*args, **kwargs)
+        if report_id is not None:
+            self.context["report_id"] = report_id
```
acroz: I'm not crazy about this use of the schema context to pass around values like this - it introduces a lot of coupling between the schemas. Why is it necessary to have the report ID in the version model?

sherlockml/clients/report.py
```diff
@@ -0,0 +1,174 @@
...
+
+from marshmallow import Schema, fields, post_load, pre_load
+
+from sherlockml.clients.base import BaseClient
+
+Report = namedtuple(
+    "Report", ["created_at", "name", "id", "description", "active_version"]
+)
+
+VersionedReport = namedtuple(
```
acroz: Can we call this (and the associated schema) something like 'ReportWithVersions'? All reports are 'versioned' so I think this name may be slightly misleading.

sherlockml/clients/report.py
```diff
@@ -0,0 +1,174 @@
...
+            self.context["report_id"] = data["report_id"]
+        return data
+
+    @post_load
+    def make_active_report(self, data):
+        return Report(**data)
+
+
+class VersionedReportSchema(Schema):
+
```
acroz: Remove empty line for consistency?

sherlockml/clients/report.py
```diff
@@ -0,0 +1,174 @@
...
+class ReportClient(BaseClient):
+
+    SERVICE_NAME = "tavern"
+
+    def list(self, project_id):
+        endpoint = "/project/{}".format(project_id)
+        return self._get(endpoint, ReportSchema(many=True))
+
+    def list_report_versions(self, report_id):
```
acroz: Maybe `get_with_versions` instead? I expect to get a list of something from a method labelled `list...`, whereas here we're getting a 'VersionedReport' back.

sherlockml/clients/report.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def create(
+        self,
+        name,
+        notebook_path,
+        project_id,
+        author_id,
+        description=None,
+        show_code=False,
+    ):
```
acroz: Our usual conventions would be to put `project_id` first - can we do that here?

acroz: (I appreciate this is probably not well established in this library - more of a general convention in out codebase)

sherlockml/clients/report.py
```diff
@@ -0,0 +1,174 @@
...
+        notebook_path = _relative_project_path(notebook_path)
```
acroz: I don't think this belongs in the client - we should just make it clear that this client only accepts paths relative to the project root.

sherlockml/clients/report.py
```diff
@@ -0,0 +1,174 @@
...
+
+        payload = {
+            "report_name": name,
+            "author_id": author_id,
+            "notebook_path": notebook_path,
+            "description": description if description is not None else "",
```
acroz: I looked at Tavern and I believe that passing `None` here (mapped to `null` in JSON) is accepted - I would recommend doing that (assuming it works against a real tavern).

sherlockml/clients/report.py
```diff
@@ -0,0 +1,174 @@
...
+        project_id,
+        author_id,
+        description=None,
+        show_code=False,
+    ):
+        notebook_path = _relative_project_path(notebook_path)
+
+        payload = {
+            "report_name": name,
+            "author_id": author_id,
```
acroz: This method should accept standard library `uuid.UUID` objects, which the standard library JSON package does not accept by default - we probably need to do `str(author_id)` here to support this.

I would suggest making all `*_id` values in the tests `uuid.UUID`s to catch this kind of issue.

sherlockml/clients/report.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def create_version(
+        self, report_id, notebook_path, author_id, show_code=False
+    ):
+
+        notebook_path = _relative_project_path(notebook_path)
+
+        payload = {
+            "notebook_path": notebook_path,
+            "author_id": author_id,
```
acroz: As above comment wrt UUIDs.

sherlockml/clients/report.py
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        endpoint = "/project/{project_id}".format(project_id=project_id)
+        return self._post(endpoint, ReportSchema(), json=payload)
+
+    def create_version(
+        self, report_id, notebook_path, author_id, show_code=False
+    ):
+
+        notebook_path = _relative_project_path(notebook_path)
```
acroz: As above comment.

sherlockml/clients/report.py
acroz: Looking at your tests I realise you're already doing this - the way our tests work does not currently catch this issue. I've added a note to make our tests catch this kind of issue in general - for now I would try to manually test the code with `uuid.UUID`s agains a real tavern to validate that there are no serialisation errors.

tests/clients/test_report.py
```diff
@@ -0,0 +1,224 @@
...
+    versions=[ACTIVE_VERSION],
+)
+
+
+REPORT_BODY = {
+    "created_at": REPORT.created_at.strftime("%Y-%m-%dT%H:%M:%S.%f")[:-3],
+    "report_name": REPORT.name,
+    "report_id": str(REPORT.id),
+    "description": REPORT.description,
+    "active_version": {
```
acroz: I would factor this inner dict out from here and from `VERSIONED_REPORT_BODY` as e.g. REPORT_VERSION_BODY` to avoid repetition, and also add a separate `test_report_version_schema` where you explicitly test is deserialisation.

sherlockml/clients/report.py
janfreyberg: This is to get around an inconsistency in what Tavern returns:

- GET report/<report_id>/versions returns a report, with a list of versions, which *don't* contain the report_id
- POST report/<report_id>/version returns just a version, which *also* contains the report_id

I think I would prefer to have the python representation of a Report Version be consistent in what it contains, hence the use of context

janfreyberg:
Sounds good - I'll drop by later!

janfreyberg:
thanks @srstevenson - sorry @acroz, I will pick this up tomororw!

janfreyberg:
@acroz addressed issues - can you take a look?

janfreyberg:
OK I think I accidentally rebased on master, hence there being so many commits in this PR.

I can probably make a new one if needed

sherlockml/clients/report.py
```diff
@@ -0,0 +1,141 @@
...
+
+class ReportSchema(Schema):
+    created_at = fields.DateTime(required=True)
+    name = fields.String(required=True, data_key="report_name")
+    id = fields.UUID(required=True, data_key="report_id")
+    description = fields.String(required=True)
+    active_version = fields.Nested(ReportVersionSchema, required=True)
+
+    @post_load
+    def make_active_report(self, data):
```
acroz:
```suggestion
    def make_report(self, data):
```

sherlockml/clients/report.py
```diff
@@ -0,0 +1,141 @@
...
+class ReportWithVersionsSchema(Schema):
+    created_at = fields.DateTime(required=True)
+    name = fields.String(required=True, data_key="report_name")
+    id = fields.UUID(required=True, data_key="report_id")
+    description = fields.String(required=True)
+    active_version_id = fields.UUID(required=True)
+    versions = fields.Nested(ReportVersionSchema, required=True, many=True)
+
+    @post_load
+    def make_versioned_report(self, data):
```
acroz:
```suggestion
    def make_report_with_versions(self, data):
```

sherlockml/clients/report.py
```diff
@@ -0,0 +1,141 @@
...
+        payload = {
+            "notebook_path": str(notebook_path),
+            "author_id": str(author_id),
+            "show_input_cells": show_code,
+            "draft": False,
+        }
+
+        endpoint = "/report/{report_id}/version".format(report_id=report_id)
+        return self._post(
+            endpoint, ReportVersionSchema(unknown=EXCLUDE), json=payload
```
acroz: I think `unknown=EXCLUDE` is the default behaviour - can we exclude this option here?

tests/clients/test_report.py
```diff
@@ -0,0 +1,213 @@
...
+    id=REPORT_ID,
+    name="Test Report Name",
+    description="Looking forward to the test reports on this Test Report",
+    created_at=datetime.datetime(2018, 10, 3, 9, 23, 0, 0),
+    active_version_id=VERSION_ID,
+    versions=[ACTIVE_VERSION],
+)
+
+VERSION_BODY = {
+    "created_at": ACTIVE_VERSION.created_at.strftime("%Y-%m-%dT%H:%M:%S.%f")[
```
acroz: Did you consider just having hard coded strings here (and other similar places)? It's probably more readable.

tests/clients/test_report.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+PROJECT_ID = uuid.uuid4()
+USER_ID = uuid.uuid4()
+REPORT_ID = uuid.uuid4()
+VERSION_ID = uuid.uuid4()
+PROJECT_ID = uuid.uuid4()
+
+ACTIVE_VERSION = ReportVersion(
+    created_at=datetime.datetime(2018, 10, 3, 9, 23, 5, 0),
```
acroz: Can we test this with datetimes with a timezone? We normally pass around zoned datetimes for explicitness, and it would be good for these tests to reflect that.

See for an example `tests/clients/test_workspace.py` line 39.

janfreyberg: I'm just trying this (it's making the tests fail because deserialised responses don't have a TZ).

So should the deserialisation assign UTC as the timezone by default so that the python object has a datetime with timezone?

tests/clients/test_report.py
janfreyberg: good point!

tests/clients/test_report.py
acroz: We expect the API to return a datetime with a timezone (a trailing Z indictating UTC as in the tests you've implemented), which marshmallow correctly interprets.

janfreyberg:
Awesome, thank you!

acroz:
This has been released in 0.18.3.

PR #33, opened 2018-10-19
Workspace client

ameroueh:


acroz:
Hey, thanks for the contribution. Before reviewing, can you fix the outstanding test failures? Looks like it's just that you have a few type hints, which don't exist on some of the older Python versions this library needs to support.

PR #34, opened 2018-10-20
Workspace client

ameroueh:


sherlockml/clients/workspace.py
```diff
@@ -0,0 +1,91 @@
...
+        elif data["type"] == FileNodeType.FILE:
+            return File(**{key: data[key] for key in File._fields})
+        else:
+            raise ValueError("Invalid file node type.")
+
+
+class ListResponseSchema(Schema):
+
+    project_id = fields.UUID(data_key="project_id", required=True)
+    path = fields.Str(required=True)
```
acroz:
```suggestion
    path = fields.String(required=True)
```

tests/clients/test_workspace.py
```diff
@@ -0,0 +1,120 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+
+from datetime import datetime
+from dateutil.tz import UTC  # type: ignore
```
acroz: Since we're not using type checking at all, I suggest not putting `type: ignore` here.

```suggestion
from dateutil.tz import UTC
```

tests/clients/test_workspace.py
```diff
@@ -0,0 +1,120 @@
...
+from uuid import uuid4
+
+from marshmallow import ValidationError
+
+
```
acroz: I suggest removing one of the empty lines here.

ameroueh: Done.

acroz:
Merged - thanks for the contribution @ameroueh !

PR #39, opened 2018-11-07
Consolidate Travis jobs to speed up tests

srstevenson:
This change allows Travis to reuse the same build host for the Python 2.7, 3.4, 3.5, and 3.6 linting and unit tests, reducing execution time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,34 +4,24 @@ install:
   - pip install flake8
 jobs:
   include:
-    - name: Check black formatting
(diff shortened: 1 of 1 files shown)


srstevenson:
The list of Python versions 2.7 to 3.6 [doesn't appear](https://travis-ci.org/ASIDataScience/sherlockml/jobs/452123503) to be used as expected: I'll investigate this, but don't review this for now.

PR #40, opened 2018-11-08
Add a new client for users

liamcoatman:
Create a new 'user' client. 

The client formerlly called 'user' has been renamed 'account'. We should check this isn't a breaking change before merging.

### DIFF
--- a/sherlockml/clients/user.py
+++ b/sherlockml/clients/user.py
@@ -50,8 +50,10 @@ class UserSchema(Schema):
     email = fields.Str(required=True)
     created_at = fields.DateTime(data_key="createdAt", required=True)
     enabled = fields.Boolean(required=True)
-    global_roles = EnumField(
-        GlobalRole, by_value=True, many=True, required=True
+    global_roles = fields.List(
+        EnumField(GlobalRole, by_value=True),
+        data_key="globalRoles",
+        required=True,
     )
 
     @post_load
--- a/sherlockml/clients/user.py
+++ b/sherlockml/clients/user.py
@@ -37,7 +37,7 @@ User = namedtuple(
         "email",
         "created_at",
         "enabled",
-        "global_roles"
+        "global_roles",
     ],
 )
 
@@ -50,7 +50,11 @@ class UserSchema(Schema):
     email = fields.Str(required=True)
     created_at = fields.DateTime(data_key="createdAt", required=True)
     enabled = fields.Boolean(required=True)
-    global_roles = fields.List(EnumField(GlobalRole, by_value=True), data_key="globalRoles", required=True)
+    global_roles = fields.List(
+        EnumField(GlobalRole, by_value=True),
+        data_key="globalRoles",
+        required=True,
+    )
 
     @post_load
     def make_project(self, data):
--- a/sherlockml/clients/user.py
+++ b/sherlockml/clients/user.py
@@ -14,10 +14,18 @@
 
 
 from collections import namedtuple
+from enum import Enum
 
 from marshmallow import Schema, fields, post_load
+from marshmallow_enum import EnumField
 
 from sherlockml.clients.base import BaseClient
+
+
+class GlobalRole(Enum):
+    BASIC_USER = "global-basic-user"
+    FULL_USER = "global-full-user"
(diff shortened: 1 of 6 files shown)


sherlockml/clients/user.py
```diff
@@ -14,38 +14,70 @@
...
+    enabled = fields.Boolean(required=True)
+    global_roles = fields.List(
+        EnumField(GlobalRole, by_value=True),
+        data_key="globalRoles",
+        required=True,
+    )
 
     @post_load
-    def make_user(self, data):
+    def make_project(self, data):
```
srstevenson: Was this renaming intentional?
```suggestion
    def make_user(self, data):
```

sherlockml/clients/user.py
```diff
@@ -14,38 +14,70 @@
...
+    SERVICE_NAME = "flock"
 
-class UserClient(BaseClient):
+    def get_user(self, user_id):
+        endpoint = "/user/{}".format(user_id)
+        response = self._get(endpoint, UserSchema())
+        return response
 
-    SERVICE_NAME = "hudson"
+    def get_users(self):
```
srstevenson: Do you think `get_all_users` is a more explicit name? (I don't feel strongly about this).
```suggestion
    def get_all_users(self):
```

tests/clients/test_user.py
```diff
@@ -11,68 +11,118 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
```
srstevenson: Can we add a blank line between the licence header and first import?

tests/clients/test_user.py
```diff
@@ -11,68 +11,118 @@
...
 import uuid
+from datetime import datetime
 
 import pytest
 from marshmallow import ValidationError
+from dateutil.tz import UTC
```
srstevenson: `from dateutil.tz import UTC` should be before `from marshmallow import ValidationError`.

tests/clients/test_user.py
```diff
@@ -11,68 +11,118 @@
...
-)
-
+from sherlockml.clients.user import UserClient, User, UserSchema, GlobalRole
 from tests.clients.fixtures import PROFILE
 
 USER_ID = uuid.uuid4()
+CREATED_AT = datetime(2018, 3, 10, 11, 32, 6, 247000, tzinfo=UTC)
+CREATED_AT_STRING = "2018-03-10T11:32:06.247Z"
+
+TEST_USER_JSON = dict(
```
srstevenson: I'd prefer we use a dictionary literal rather than calling `dict` (using a literal is a bit faster, as the interpreter doesn't need to lookup `dict`).

sherlockml/clients/user.py
```diff
@@ -14,38 +14,70 @@
...
 
 class UserSchema(Schema):
+
     id = fields.UUID(data_key="userId", required=True)
+    username = fields.Str(required=True)
+    full_name = fields.Str(data_key="fullName", required=True)
+    email = fields.Str(required=True)
+    created_at = fields.DateTime(data_key="createdAt", required=True)
+    enabled = fields.Boolean(required=True)
+    global_roles = fields.List(
```
srstevenson: Would `global_roles` be better as a set rather than a list?

liamcoatman: I don't think marhmellow supports a set type

srstevenson: Ah OK, let's leave it as a list then.

zblz:
@acroz - You have more context on this. AFAICT, we were not using the `UserClient.get_authenticated_user_id` anywhere yet, so it's renaming to `AccountClient` will not have any side effects. Is that true?

sherlockml/clients/user.py
acroz: I believe our existing convention is to use `many=True` on the inner item's schema rather than putting it in a `fields.List`. I like the explicitness of the latter, but the former is more compact. Should we adopt a consistent style here?

acroz:
These changes now released in 0.18.0

PR #41, opened 2018-11-13
Add job and log clients

acroz:


sherlockml/clients/job.py
```diff
@@ -0,0 +1,303 @@
...
+
+class JobClient(BaseClient):
+
+    SERVICE_NAME = "steve"
+
+    def list(self, project_id):
+        endpoint = "/project/{}/job".format(project_id)
+        return self._get(endpoint, JobSummarySchema(many=True))
+
+    def create_run(self, project_id, job_id, parameter_value_sets):
```
pbugnion: Should there be a docstring for creating a run? In particular, the structure of `parameter_value_sets` is not obvious.

pbugnion: Not sure if it's conventional in this, but should we consider a default `parameter_value_sets=None` (obviously defaulting to `[{}]`)?

pbugnion:
Should the client for job logs be called just `LogClient`? Feels like it precludes adding log clients for other things.

pbugnion:
Looks good, modulo some comments inline.

acroz:
> Should the client for job logs be called just `LogClient`? Feels like it precludes adding log clients for other things.

I discussed this with others, and we came to the agreement that it was likely that other centralised log storage in the future would more likely be with Wozniak than not, and that calling this simply `LogClient` was suitable.

PR #43, opened 2018-11-19
Remove EnumField in favour of validators

zblz:
We've been using the `marshmallow_enum` package in a few places in the library, but it's not great: `EnumField` does not have a `many` kwarg, so it makes enum fields qualitatively different than other fields in terms of how we write them. On the other hand, we are not actually using the enumerations that we deserialise anywhere where it matters that they are an `Enum`. marshmallow used to have a `fields.Enum`, but it was deprecated in favour of using the `OneOf` and `ContainsOnly `validators, which take lists of allowed values. This PR drops the dependency on `marshmallow_enum` while keeping behaviour through validators.


acroz:
Can we keep the actual Enum types rather than reverting to strings? I think it's valuable to have a well defined set of values in an Enum rather than users of the clients having to do string comparisons. The main reason we've been using enums in my mind has been that rather than as a prerequisite for EnumField.

janfreyberg:
My vote would go for @zblz 's current implementation with sets! (I love sets)

sherlockml/clients/server.py
```diff
@@ -92,7 +83,10 @@ class ServerSchema(Schema):
     )
     server_size = fields.Nested(ServerSizeSchema, data_key="instanceSize")
     created_at = fields.DateTime(data_key="createdAt", required=True)
-    status = EnumField(ServerStatus, by_value=True, required=True)
+    status = fields.String(
+        required=True,
+        validate=validate.OneOf(["creating", "running", "error", "destroyed"]),
```
srstevenson: I have a mild preference for moving this list out into a module constant for parity with `sherlockml/clients/job.py`.

sherlockml/clients/user.py
```diff
@@ -17,17 +17,11 @@
 from enum import Enum
 
 from marshmallow import Schema, fields, post_load
-from marshmallow_enum import EnumField
+from marshmallow import validate
```
srstevenson: This can be coalesced with the imports from `marshmallow` on the previous line.

sherlockml/clients/user.py
```diff
@@ -51,9 +45,12 @@ class UserSchema(Schema):
     created_at = fields.DateTime(data_key="createdAt", required=True)
     enabled = fields.Boolean(required=True)
     global_roles = fields.List(
-        EnumField(GlobalRole, by_value=True),
+        fields.String,
         data_key="globalRoles",
         required=True,
+        validate=validate.ContainsOnly(
+            ["global-basic-user", "global-full-user", "global-admin"]
```
srstevenson: Ditto about using a module constant.

acroz:
@janfreyberg - sets are great, though I think the question of whether to use strings or enums is a somewhat separate question.

zblz:
I feel like using `Enum`s is a good idea in general but in this case I'm not sure we'll be using them for anything beyond validation and display purposes, i.e., users of the client are unlike to use the values of the Enums directly.

If we want to keep the enums then it's probably better to keep using the `marshmallow_enum` library, otherwise we actually need to convert from validated strings into Enum members in the `@post_load` functions, which means we would add extra complexity instead of reducing it.

acroz:
I discussed this with @zblz and we agreed to keep these fields as `Enum`s, and therefore we should also keep the `EnumField` from `marshmallow_enum`. We may want to submit a PR to `marshmallow_enum` to support `any=True`.

PR #45, opened 2018-11-29
Update SherlockMLAuth docstring.

sbalian:
Just added `services` in the two URLs.

### DIFF
--- a/sherlockml/clients/auth.py
+++ b/sherlockml/clients/auth.py
@@ -71,7 +71,7 @@ class SherlockMLAuth(AuthBase):
     ``requests``:
 
     >>> import requests
-    >>> requests.get('https://service.services.example.sherlockml.net',
+    >>> requests.get('https://servicename.services.example.sherlockml.net',
                      auth=auth)
 
     You can also set it as the ``auth`` attribute on a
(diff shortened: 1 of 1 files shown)


sherlockml/clients/auth.py
```diff
@@ -64,14 +64,15 @@ class SherlockMLAuth(AuthBase):
...
-    >>> auth = SherlockMLAuth('https://hudson.example.sherlockml.net',
+    >>> auth = SherlockMLAuth('https://hudson.services.example.sherlockml.net',
                               your_client_id, your_client_secret)
 
     then pass it as the ``auth`` argument when making a request with
     ``requests``:
 
     >>> import requests
-    >>> requests.get('http://service.example.sherlockml.net', auth=auth)
+    >>> requests.get('https://service.services.example.sherlockml.net',
```
janfreyberg: can we replace this with `https://servicename.services.example.sherlockml.net` to make it less ambiguous?

acroz: :+1:

sbalian: Changed.

janfreyberg:
Nice! Good to merge.

sbalian:
Thanks @janfreyberg . Can you merge it? (I'm not authorised for some reason)

janfreyberg:
Me neither! I think only @acroz is

PR #47, opened 2018-12-12
Support Faculty-prefixed environment variables

acroz:
This PR adds support for configuring with new `FACULTY_`-prefixed environment variables, and tests that these have precedence over the preexisting `SHERLOCKML_`-prefixed env vars.

Support for the new variables has not yet been added to the datasets module.

### DIFF
--- a/sherlockml/config.py
+++ b/sherlockml/config.py
@@ -73,18 +73,23 @@ def resolve_profile(
 
     resolved_credentials_path = (
         credentials_path
+        or os.getenv("FACULTY_CREDENTIALS_PATH")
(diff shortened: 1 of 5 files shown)


pbugnion:
What do we think `FACULTY_CREDENTIALS_PATH` is going to be?

srstevenson:
> What do we think `FACULTY_CREDENTIALS_PATH` is going to be?

`${XDG_CONFIG_HOME:-$HOME/.config}/faculty/credentials`.

pbugnion:
Are you planning on doing datasets and `SHERLOCKML_PROJECT_ID`?

PR #48, opened 2018-12-13
Rename to faculty

pbugnion:
This PR renames the module structure and the classes to faculty. In particular, this introduces the following work for users:

- rename any import from `sherlockml` to `faculty`.
- replace the very outdated `sherlockml.filesystem` with `sherlockml.datasets`
- `SherlockMLAuth` is deprecated in favour of `FacultyAuth`, but it still works.

Explicitly not handled in this PR:
- the credentials are still stored in `$XDG_CONFIG_HOME/sherlockml/credentials`. The plan is to do this in a separate PR.


faculty/clients/auth.py
```diff
@@ -104,3 +105,12 @@ def __call__(self, request):
         request.headers["Authorization"] = header_content
 
         return request
+
+
+class SherlockMLAuth(FacultyAuth):
```
acroz: I'm not sure that we want to have this alias here at all - if users are already migrating from using the `sherlockml` library to `faculty`, it seems strange to allow them to leave this `Auth` class as `SherlockMLAuth` and not just force them to update that as well.

Spoke with @pbugnion about this and he thinks there's some value in allowing people to upgrade gradually. Any thoughts @zblz @srstevenson ?

srstevenson: I side with @acroz: I favour replacing all references to SherlockML with references to Faculty atomically. We're changing the package name and import paths, so users will already need to change their imports and can change the class name at the same time. A `grep -i` for `sherlockml` will allow them to ensure they've not missed anything.

zblz: I think `SherlockMLAuth(FacultyAuth)` should only be in the `sherlockml` library during the deprecation period. As @acroz and @srstevenson said, the `faculty` library should be free of SherlockML references (except maybe using old env vars and paths for credentials)

README.rst
```diff
@@ -4,4 +4,4 @@ sherlockml
 .. image:: https://travis-ci.org/ASIDataScience/sherlockml.svg?branch=master
     :target: https://travis-ci.org/ASIDataScience/sherlockml
```
zblz: This and the line above should point at the faculty Travis project.

faculty/clients/auth.py
```diff
@@ -49,29 +50,29 @@ def get_access_token(self, client_id, client_secret):
...
-    >>> auth = SherlockMLAuth('https://hudson.services.example.sherlockml.net',
-                              your_client_id, your_client_secret)
+    >>> auth = FacultyAuth('https://hudson.services.example.my.faculty.ai',
+                           your_client_id, your_client_secret)
 
     then pass it as the ``auth`` argument when making a request with
     ``requests``:
 
     >>> import requests
-    >>> requests.get('https://servicename.services.example.sherlockml.net',
```
zblz: `s/example.faculty.ai/example.my.faculty.ai/`

faculty/clients/auth.py
```diff
@@ -104,3 +105,12 @@ def __call__(self, request):
...
         return request
+
+
+class SherlockMLAuth(FacultyAuth):
+    def __init__(self, *args, **kwargs):
+        warnings.warn(
+            "SherlockMLAuth has been renamed FacultyAuth - please update "
+            "your code to match the new class name."
+        )
+        super().__init__(*args, **kwargs)
```
zblz: Also, this is not Python2 compatible, but it goes away if we drop `SherlockMLAuth`

faculty/clients/auth.py
pbugnion: Fair enough. I'll remove it.

README.rst
pbugnion: good catch.

pbugnion:
I have tested this on a sherlock server (specifically the server client) and it seems to work.

pbugnion:
In terms of migration plan, once this is merged I'm going to push a version of current sherlockml onto bitbucket, as discussed. We will then install both sherlockml and faculty into user images.

pbugnion:
The legacy sherlockml repository is now available on our private bitbucket as `sherlockml`.

PR #50, opened 2019-01-15
Fold in sml

mociarain:
An initial addition of the faculty-cli (sml) to the faculty library. This is just a wholesale copy and paste with some s/sherlock/faculty for good measure. i.e.

sml projects --> faculty projects

### DIFF
--- a/faculty/cli/cli.py
+++ b/faculty/cli/cli.py
@@ -980,7 +980,7 @@ def job_run_logs(project, job, run):
 
     project_id, job_id = _resolve_job(project, job)
 
-    job_client = faculty.client("job") 
+    job_client = faculty.client("job")
     run_details = job_client.get_run(project_id, job_id, run.run_number)
     if run.subrun_number is not None:
         subrun_number = run.subrun_number
--- a/faculty/cli/cli.py
+++ b/faculty/cli/cli.py
@@ -32,21 +32,21 @@ from distutils.version import StrictVersion
 
 import click
 import requests
-import faculty.config
-import faculty.cli.auth
 from tabulate import tabulate
 
 import faculty.cli.auth
-import faculty.cli.config
 import faculty.cli.baskerville
 import faculty.cli.casebook
 import faculty.cli.client
+import faculty.cli.config
 import faculty.cli.galleon
 import faculty.cli.hound
 import faculty.cli.parse
 import faculty.cli.shell
 import faculty.cli.update
 import faculty.cli.version
+import faculty.config
+from faculty.clients.base import NotFound
 
 SSH_OPTIONS = [
     "-o",
@@ -72,7 +72,7 @@ def _print_and_exit(msg, code):
 
 def _get_pypi_versions():
     """List releases available from PyPI."""
-    ## TODO needs to point to the new repo
+    # TODO needs to point to the new repo
     response = requests.get("https://pypi.python.org/pypi/sml/json", timeout=1)
     versions = response.json()["releases"].keys()
     return [StrictVersion(v) for v in versions]
@@ -146,7 +146,7 @@ def _ensure_creds_file_present():
     except IOError:
         msg = textwrap.dedent(
             """\
-        It looks like this is the first time you've used the faculty cli on 
+        It looks like this is the first time you've used the faculty cli on
         this computer, so you must enter your Faculty credentials. They'll
         be saved so you don't have to enter them again.
         """
--- a/faculty/cli/cli.py
+++ b/faculty/cli/cli.py
@@ -1,0 +1,1193 @@
+"""Command line interface."""
+
+# Copyright 2016-2018 ASI Data Science
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import division
+
+import contextlib
+import operator
+import os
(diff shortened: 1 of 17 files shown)


srstevenson:
This is -1 from me: `sml` is legacy, untested code, which wasn't written for third party use. Additionally, it implements clients for all the Faculty services itself rather than using the tested clients in the `faculty` library. We should implement a new CLI with good test coverage and treating Windows as a supported platform, using the `faculty` library.

mociarain:
Sorry I should have been clearer for people outside the engineering team. This is meant as an interim solution to get a faculty CLI. The duplication and lack of tests is not meant to be permanent or even that temporary. As for starting from scratch ¯\_(ツ)_/¯

faculty/cli/config.py
```diff
@@ -0,0 +1,77 @@
...
+            fg="yellow",
+        )
+    return profile_name
+
+
+def get_profile():
+    global _PROFILE_CACHE
+    if _PROFILE_CACHE is None:
+        profile_name = (
+            os.getenv("FACULTY_PROFILE")
```
acroz: If support for `*_ENV` above is dropped, we won't need to check `*_PROFILE` here as we can just rely on `faculty.config.resolve_profile` to do that.

faculty/cli/config.py
```diff
@@ -0,0 +1,77 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import os
+
+import click
+import faculty.config
```
acroz: We usually put an empty line between third party (`click`) and package-internal (`faculty.config`) imports.

faculty/cli/client.py
```diff
@@ -0,0 +1,169 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+from contextlib import contextmanager
+
+import requests
+import faculty.cli.auth
```
acroz: We usually put an empty line between third party (`click`) and package-internal (`faculty.config`) imports.

faculty/cli/cli.py
```diff
@@ -0,0 +1,1193 @@
...
+def _print_and_exit(msg, code):
+    """Print error message and exit with given code."""
+    click.echo(msg, err=True)
+    sys.exit(code)
+
+
+def _get_pypi_versions():
+    """List releases available from PyPI."""
+    # TODO needs to point to the new repo
+    response = requests.get("https://pypi.python.org/pypi/sml/json", timeout=1)
```
acroz:
```suggestion
    response = requests.get("https://pypi.org/pypi/faculty/json", timeout=1)
```

acroz: (Can also then remove the comment above)

faculty/cli/cli.py
```diff
@@ -0,0 +1,1193 @@
...
+    versions = response.json()["releases"].keys()
+    return [StrictVersion(v) for v in versions]
+
+
+def _populate_creds_file():
+    """Prompt user for client ID and secret and save them."""
+    while True:
+        domain = click.prompt(
+            "Domain", default="services.sherlockml.com", err=True
```
acroz: Update this to `services.faculty.ai` (or whatever the services domain for cloud is going to be)?

faculty/cli/version.py
```diff
@@ -0,0 +1,32 @@
...
+try:
+    __version__ = get_distribution("faculty").version
+except DistributionNotFound:
+    # package is not installed
+    pass
+
+
+def user_agent():
+    """Return user agent string."""
+    return "sml/{} (Python {}, {})".format(
```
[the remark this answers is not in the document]
srstevenson: `faculty` sounds good to me.

faculty/cli/update.py
```diff
@@ -0,0 +1,85 @@
...
+
+def _check_for_new_release():
+    current = StrictVersion(faculty.cli.version.__version__)
+    latest = max(_get_pypi_versions())
+    if current < latest:
+        template = (
+            "You are using faculty version {}, however version {} is "
+            "available.\n"
+            "You should upgrade with 'pip install --upgrade faculty'."
+        )
```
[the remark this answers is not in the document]
srstevenson: I'm against folding `sml` into this library, so prefer having a separate `faculty-cli` package as a long term solution. As Andrew says, this avoids CLI users being told to upgrade when there is no functional change to the CLI, and additionally avoids library users from having an additional executable placed on their path when they're only using the Python API.

faculty/cli/config.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+_PROFILE_CACHE = None
+
+
+def _profile_name_from_faculty_env():
+    profile_name = os.getenv("FACULTY_ENV") or os.getenv("SHERLOCKML_ENV")
+    if profile_name is not None:
+        click.secho(
+            "FACULTY_ENV is deprecated and will be removed in a future "
```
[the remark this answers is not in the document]
srstevenson: We've never used `FACULTY_ENV` before, so it doesn't seem worthwhile to add it here as it'll be deprecated from inception.

faculty/cli/cli.py
srstevenson: `services.cloud.my.faculty.ai` says @neb42.

faculty/cli/config.py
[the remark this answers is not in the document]
zblz: Indeed, I think the name change is a great moment to drop legacy stuff.

faculty/cli/update.py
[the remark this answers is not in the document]
zblz: I was initially in favour of folding the CLI into `faculty`, but I'm coming around to a separate package being a better idea.

mociarain:
This was closed as we no longer want to integrate the cli directly in this library. It has been moved [faculty-cli](https://pypi.org/project/faculty-cli/).

PR #51, opened 2019-01-16
Use new casebook client

mociarain:
This PR removes the Casebook brought in with the cli and uses the existing ProjectClient.

I also added tests for the effected cli commands and added a tests/fixtures.py. I haven't combined the existing tests/clients/fixtures.py into this but I would be in favour of doing this unless someone has a good reason not too.


mociarain:
See: https://github.com/ASIDataScience/faculty/pull/50

PR #58, opened 2019-01-22
Refactor auth module to have module-level cache

acroz:
A future PR will aim to add disk-based caching of auth tokens too.

PR #59, opened 2019-01-22
Add session

acroz:
Extending on the auth refactoring @mociarain and I did earlier, I've implemented a Session class which:

* Keeps a cache of access tokens
* Provides a helper to construct URLs for service endpoints
* Implements a `.get()` classmethod which loads the profile and caches sessions (and consequently caches the profile and access tokens)

`FacultyAuth`, the existing clients and the high level `faculty.client` interface have been updated to use this new session class.

This supersedes PR #58.


faculty/clients/base.py
```diff
@@ -127,27 +120,23 @@ class BaseClient(object):
...
         if self.SERVICE_NAME is None:
             raise RuntimeError(
                 "must set SERVICE_NAME in subclasses of BaseClient"
             )
-        self.profile = profile
+        self.session = session
         self._http_session_cache = None
 
     @property
     def http_session(self):
```
zblz: Now that we have a long-lived session object, I think it also makes sense to move the requests session there. This way the client base doesn't even need to know anything about authentication, it's all managed at the session level.

faculty/session.py
```diff
@@ -0,0 +1,99 @@
...
+    token = body["access_token"]
+    now = datetime.now(tz=pytz.utc)
+    expires_at = now + timedelta(seconds=body["expires_in"])
+
+    return AccessToken(token, expires_at)
+
+
+class Session(object):
+
+    __session_cache = {}
```
zblz: Why is this an attribute of the class? What happens when multiple Sessions are instantiated for different profiles, do their caches point at the same object or is a new cache attached to each object? I feel it would be simpler to put it at the module level.

faculty/session.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def __init__(self, profile, access_token_cache):
+        self.profile = profile
+        self.access_token_cache = access_token_cache
+
+    @classmethod
+    def get(cls, access_token_cache=None, **kwargs):
```
zblz: I've mentioned it above, but I'm not sure what the advantage is of having a classmethod to get a Session vs a module-level `get_session` function that retrieves sessions from a module-level cache or instantiates them as necessary. I like the `faculty.session.get_session()` route better than `faculty.session.Session.get()` considering you might be getting an existing session.

faculty/session.py
```diff
@@ -0,0 +1,99 @@
...
+        key = tuple(kwargs.items()) + (access_token_cache,)
```
zblz: We are not considering the (admittedly unlikely) case where env vars or config files change between invocations of `get`, so that for the same `kwargs` of get you would get different profiles and therefore different sessions. I know there is a cost in calling `resolve_profile`, but it might be worth keying the cache on the resolved profile to make sure we are returning the correct session.

faculty/session.py
```diff
@@ -0,0 +1,99 @@
...
+        if access_token is None or access_token.expires_at < utc_now:
+            return None
+        else:
+            return access_token
+
+    def add(self, profile, access_token):
+        self._store[profile] = access_token
+
+
+def _service_url(profile, service, endpoint=""):
```
zblz: `_service_url` and `_get_access_token` are only used AFAICT from within `Session`. Is there a reason not to put them as static methods of the class?

tests/test_session.py
```diff
@@ -0,0 +1,176 @@
...
+    }
+    assert access_token == AccessToken(
+        token="access-token", expires_at=IN_TEN_MINUTES
+    )
+
+
+def test_session_get(mocker, isolated_session_cache):
+    mocker.patch("faculty.config.resolve_profile", return_value=PROFILE)
+    access_token_cache = mocker.Mock()
+    mocker.spy(Session, "__init__")
```
zblz: Nice, didn't know about spy!

faculty/clients/auth.py
```diff
@@ -12,66 +12,34 @@
...
+    session : faculty.session.Session
+        The Faculty session to authenticate with
 
     To perform an authenticated request against a Faculty service, first
     construct an instance of this class:
 
-    >>> auth = FacultyAuth('https://hudson.services.example.my.faculty.ai',
-                           your_client_id, your_client_secret)
+    >>> from faculty.session import Session
+    >>> from faculty.clients.auth import FacultyAuth
```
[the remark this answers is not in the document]
acroz: Yeah, fair, but I wanted to use `faculty.session.Session` without explaining where that came from (so put in the import) and then having that import and not one for `FacultyAuth` seemed weird. If you think it's fine to omit it though I'm happy to do so.

faculty/__init__.py
```diff
@@ -13,11 +13,11 @@
...
 
-import faculty.config
+import faculty.session
 import faculty.clients
 
 
-def client(resource, *args, **kwargs):
-    profile = faculty.config.resolve_profile(*args, **kwargs)
+def client(resource, **kwargs):
+    session = faculty.session.Session.get(**kwargs)
```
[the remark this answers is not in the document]
acroz: Well, this still has 2 mentions of `session`, just in a different order.

The main motivation for having the `get` helper be a classmethod was that we had envisioned a pattern on the resource-based interface where classmethods would help you get objects, e.g.:

```python
project = Project.get(name="my project")
```

@pbugnion suggested and I agreed that it was good to follow a similar pattern here for the sake of consistency.

faculty/clients/base.py
acroz: Perhaps, since this was mostly an implementation detail of the `BaseClient`, I was planning on deferring this to a future PR.

There is a little complexity to be introduced by needing a slightly different `Auth` class for sending the token as a cookie instead of an HTTP header (for the hound client), and since that need to 'choose' an auth type is client-specific, I think it makes sense to wait until we have that before doing further refactoring.

faculty/session.py
acroz: I think if the factory method is on the class, the cache should be too, but maybe let's put off discussion of that until the former question is decided.

faculty/session.py
acroz: The latter point is an interesting different from the resource interface use case I described in your earlier comment. Let's discuss on that other comment to keep it all in the same place.

faculty/session.py
acroz: Yes, interesting point. @pbugnion do you have any thoughts on this (we had discussed it yesterday).

tests/test_session.py
acroz: 👍

faculty/session.py
acroz: I generally prefer putting less stuff in the class as it's easier to reason about dependencies and thinking about tests when they're obviously not dependent on the class itself.

faculty/__init__.py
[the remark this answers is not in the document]
zblz: > Well, this still has 2 mentions of `session`, just in a different order.

It could also be `faculty.session.get()`, but `from session import get` would leave an orphan get without much meaning.
 
> The main motivation for having the `get` helper be a classmethod was that we had envisioned a pattern on the resource-based interface where classmethods would help you get objects, e.g.:
> 
> ```python
> project = Project.get(name="my project")
> ```
> 
> @pbugnion suggested and I agreed that it was good to follow a similar pattern here for the sake of consistency.

Yeah, in that model this definitely makes sense. However, we'd probably put the resource classes at the top level of the package, so it would be `faculty.Project`. Should we do the same with `Session` (import it in `faculty/__init__.py`)?

faculty/session.py
acroz: Discussed this with @pbugnion and we decided to stick with the current behaviour.

faculty/clients/auth.py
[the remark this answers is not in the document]
acroz: Ok, I've satisfied myself that not importing `FacultyAuth` here is fine.

zblz:
LGTM, thanks for the changes. The only optimization we could think about is whether getting a session is important enough to be imported at the top of the package: i.e. `faculty.get_session()` as opposed to `faculty.session.get_session()`. I consider it is as it is a prerequisite to using the rest of the library, but I this is minor compared to the rest of the changes in the package.

acroz:
> LGTM, thanks for the changes. The only optimization we could think about is whether getting a session is important enough to be imported at the top of the package: i.e. `faculty.get_session()` as opposed to `faculty.session.get_session()`. I consider it is as it is a prerequisite to using the rest of the library, but I this is minor compared to the rest of the changes in the package.

Yes, this is a good point, however I've deliberately avoided putting high level API stuff in PRs so far (apart from `faculty.client()`), to give us the freedom to design that as a whole later.

PR #60, opened 2019-01-23
Use tox

zblz:
This adds a minimal tox setup to make it easy to run CI like tests locally. It also reduces the places where the test command is specified to a single one in `tox.ini` as opposed to one for every job in `.travis.yml`.

The last line also means we could get rid of `pytest-runner` in favour of running plain `py.test` within tox. Since tox uses pip instead of setuptools to resolve dependencies, it would result in the installed packages being closer to what a user will see in an installed version of `faculty`.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,6 +4,7 @@ toxworkdir = {env:TOX_WORK_DIR:.tox}
 
 [testenv]
 sitepackages = False
+passenv = AWS_*
 commands = python setup.py test {posargs}
 
 [testenv:flake8]
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,22 @@
+[tox]
+envlist = py27, py34, py35, py36, py37, flake8, black
+toxworkdir = {env:TOX_WORK_DIR:.tox}
+
+[testenv]
+sitepackages = False
+commands = python setup.py test {posargs}
+
+[testenv:flake8]
+skip_install = True
+deps =
+    flake8
+commands =
+    flake8
+
+[testenv:black]
+skip_install = True
+basepython = python3.7
+deps =
+    black==18.9b0
+commands =
+    black {posargs:--check setup.py faculty tests}
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,5 @@
 [tox]
 envlist = py27, py34, py35, py36, py37, flake8, black
-toxworkdir = {env:TOX_WORK_DIR:.tox}
 
 [testenv]
 sitepackages = False
--- a/tox.ini
+++ b/tox.ini
@@ -15,7 +15,6 @@ commands =
 
 [testenv:black]
 skip_install = True
-basepython = python3.7
 deps =
     black==18.9b0
 commands =
--- a/tox.ini
+++ b/tox.ini
@@ -5,7 +5,12 @@ toxworkdir = {env:TOX_WORK_DIR:.tox}
 [testenv]
 sitepackages = False
 passenv = AWS_*
-commands = python setup.py test {posargs}
+deps =
+    pytest
+    pytest-mock
+    requests_mock
+    python-dateutil>=2.7
+commands = pytest {posargs}
 
 [testenv:flake8]
 skip_install = True
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ dist/
 .pytest_cache/
 __pycache__/
 *.pyc
+.tox
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,14 +4,11 @@ dist: xenial
 
 matrix:
   include:
-    - name: Check black formatting
-      python: 3.7
+    - python: 3.7
       env: TOXENV=black
-    - name: Check flake8 on Python 3
-      python: 3.7
+    - python: 3.7
       env: TOXENV=flake8
-    - name: Check flake8 on Python 2
-      python: 2.7
+    - python: 2.7
       env: TOXENV=flake8
 
     - stage: test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,39 +1,36 @@
 language: python
 cache: pip
-install:
-  - pip install flake8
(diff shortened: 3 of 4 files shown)


tox.ini
```diff
@@ -0,0 +1,21 @@
...
+skip_install = True
+deps =
+    flake8
+commands =
+    flake8
+
+[testenv:black]
+skip_install = True
+deps =
+    black==18.9b0
```
acroz: Why is this pinned to a specific black version?

acroz:
I saw that you removed the names from linting jobs - why was that? I found those names to be very helpful.

tox.ini
srstevenson: Black's formatting rules are still being iterated on between releases, so this stops an unchanged file that passed CI when it was merged failing CI later because a newer version of Black wants to reformat it (which we've experienced previously).

acroz:
If we use tox or nox, when running tests locally, does it make a new virtual environment every time?

This is only slightly related, but I'd like to maintain the ability to quickly run tests in my development virtual environment while having test dependencies maintained for me, if possible.

acroz:
> I support having tooling that makes it easy to run the whole suite of tests and linters locally as well as in CI. Have you considered [nox](http://nox.thea.codes/en/stable/) instead of tox? It's a replacement that is configured in Python, rather than tox's slightly perverse ini format. [Here's an example](https://github.com/srstevenson/slack-purger/blob/a381ce3d69ea55c9400789dffdf66d6865b0ebfe/noxfile.py) of the nox configuration in one of my personal projects.

nox looks really nice - I support looking into it.

srstevenson:
> If we use tox or nox, when running tests locally, does it make a new virtual environment every time?

I don't know about tox, but nox does by default to ensure all your test dependencies are declared. You can change this by passing `-r`/`--reuse-existing-virtualenvs` though.

tox.ini
zblz: Exactly, increasing this version should be part of a PR that also reformats the code to that version. This is also a good reason to tell developers to always use black through tox: you don't have to also specify the version, and you can format your code by running `tox -e black -- .`.

zblz:
> I saw that you removed the names from linting jobs - why was that? I found those names to be very helpful.

I found them confusing when figuring out which version of python was being used for the flake8 and black jobs. Can bring them back now that those are working.

zblz:
> I don't know about tox, but nox does by default to ensure all your test dependencies are declared. You can change this by passing `-r`/`--reuse-existing-virtualenvs` though.

Confusingly, in tox it is the other way around: they are reused by default and you need to pass `-r, --recreate` to recreate them.

zblz:
> I support having tooling that makes it easy to run the whole suite of tests and linters locally as well as in CI. Have you considered [nox](http://nox.thea.codes/en/stable/) instead of tox? It's a replacement that is configured in Python, rather than tox's slightly perverse ini format. [Here's an example](https://github.com/srstevenson/slack-purger/blob/a381ce3d69ea55c9400789dffdf66d6865b0ebfe/noxfile.py) of the nox configuration in one of my personal projects.

nox looks good, even though I've always found python as config files not as discoverable (or copyable) as simpler formats. On the other hand, you are right that tox's ini is a bit weird. 

The redeeming feature of tox is it's wider adoption and that people know how to use it, but we could definitely use nox for this project.

zblz:
I've been looking at `nox` and I feel that for simple setups `tox` is more straightforward and easier to setup. However, I see how it can enable flexible complex setups impossible in `tox`. My vote would be for using `tox` in `faculty` as its test suite is very straightforward, but I'm willing to be convinced otherwise.

tox.ini
acroz: > you can format your code by running `tox -e black -- .`

How convenient. I see the value in pinning, though.

acroz:
I've been playing around with this to try and figure out the easiest way of running tests as fits my current workflow. My main pain point at the moment is managing test dependencies. The principal advantage of using pytest-runner is that all dependencies for the package itself, for setup and for testing are specified together in setup.py.

Moving to tox invoking pytest directly seems to imply one of:
* Test dependencies are declared in tox.ini and tests must always be run through tox and I have to run e.g. `tox -e py37 -- -v -k clients` (a bit cumbersome).
* Test dependencies are in `requirements-test.txt` while package dependencies remain in `setup.py` (untidy, and requires the usuer to manually install/update dependencies).
* Test dependencies are duplicated between `setup.py` and `tox.ini` (duplication is bad due to people making mistakes).

This PR shows effectively how CI can be set up to run tox tests and I certainly appreciate the value of being able to reproduce the full test suite on all versions locally, but I'm concerned that we'll rarely actually do that, and that the 95% case of running tests regularly while developing will be much more painful. Can one of you help me out in figuring out what the development workflow will look like if tox were adopted?

zblz:
> The principal advantage of using pytest-runner is that all dependencies for the package itself, for setup and for testing are specified together in setup.py.

And I don't think that's necessarily a good thing. The main purpose of `setup.py` is packaging, and it is an important enough purpose that it should be touched as little as possible. With the `pytest-runner` setup you are also not testing packaging, so if you are missing files in your `MANIFEST.in` you'll only notice after puching to PyPI. 

> * Test dependencies are declared in tox.ini and tests must always be run through tox and I have to run e.g. `tox -e py37 -- -v -k clients` (a bit cumbersome).

The equivalent now is `python setup.py test --addopts "-v -k clients"` which is not less cumbersome. (if anything, it forces you to remember the non-pytest `--addopts` flag which I always have to look up).

zblz:
> Moving to tox (invoking pytest directly) seems to imply one of:
> 
> * Test dependencies are declared in tox.ini and tests must always be run through tox and I have to run e.g. `tox -e py37 -- -v -k clients` (a bit cumbersome).

I've added another commit that implements this move to calling pytest directly from within tox so we can see how it would look. Happy to revert and postpone that decision if you think either of the other options would be better.

PR #61, opened 2019-01-24
Add disk-persisted access token cache

acroz:
With this cache implementation, users will be able to construct clients which persist the access token cache to disk. To use:

```python
import faculty
from faculty.session import FileSystemAccessTokenCache

cache = FileSystemAccessTokenCache("~/.cache/faculty/access-tokens")
project_client = faculty.client("project", access_token_cache=cache)
```

Or if you want to manage the session yourself:

```python
from faculty.session import get_session, FileSystemAccessTokenCache

cache = FileSystemAccessTokenCache("~/.cache/faculty/access-tokens")
session = get_session(access_token_cache=cache)
```

Comments are welcome in particular on:

* The name of this class
* Should there be a default cache location at `$XDG_CACHE_DIR/faculty/token-cache.json`?
* `FileSystemAccessTokenCache` could extend `MemoryCache` and call its `get` and `add` using `super()` rather than reimplementing them as done here - any thoughts on that?
* I am considering moving the access token caches into their own module. Should that be done, and if so, should that be in a `faculty.session` subpackage, with the current `faculty.session` module broken up into component modules, or somewhere else?


faculty/session.py
```diff
@@ -26,20 +30,116 @@
...
-        utc_now = datetime.now(tz=pytz.utc)
-        if access_token is None or access_token.expires_at < utc_now:
-            return None
+        return access_token if _is_valid_access_token(access_token) else None
+
+    def add(self, profile, access_token):
+        self._store[profile] = access_token
+
+
+def _ensure_directory_exists(path, mode):
```
liamcoatman: Would something like `if not (os.path.exists(path) and os.path.isdir(path))...` be simpler here?

srstevenson:
Regarding the name, I'd prefer `AccessTokenFileSystemCache` over `FileSystemAccessTokenCache`. I think it makes it clearer that it's an access token cache on the file system, rather than a cache for file system access tokens. We could also consider abbreviating `FileSystem` to FS, as the class name is quite long and it's a commonly used acronym (c.f. AFS, HDFS, NFS, tmpfs).

faculty/session.py
```diff
@@ -26,20 +30,116 @@
...
+class AccessTokenSchema(Schema):
+    token = fields.String(required=True)
+    expires_at = fields.DateTime(data_key="expiresAt", required=True)
+
+    @post_load
+    def make_access_token(self, data):
+        return AccessToken(**data)
+
+
+class AccessTokenStore:
```
srstevenson:
```suggestion
class AccessTokenStore(object):
```

faculty/session.py
```diff
@@ -26,20 +30,116 @@
...
+    def add(self, profile, access_token):
+        self._store[profile] = access_token
+
+
+def _ensure_directory_exists(path, mode):
+    try:
+        os.makedirs(path, mode=mode)
+    except OSError as e:
+        if e.errno == errno.EEXIST:
+            # Directory already exists
```
srstevenson: I'm not sure this comment is necessary, given the use of a named constant in the comparison to `e.errno`.

faculty/session.py
```diff
@@ -26,20 +30,116 @@
...
+                data = json.load(fp)
+            self._store = AccessTokenStoreSchema().load(data)
+        except IOError as e:
+            if e.errno == errno.ENOENT:
+                # File does not exist - initialise empty store
+                self._store = AccessTokenStore()
+            else:
+                raise
+        except (ValueError, ValidationError):
+            # File is of invalid format - reset with empty store
```
srstevenson: Although we don't do any logging in the library yet (and needn't do so in this PR), in future I think we should introduce debug level logging for things like this.

faculty/session.py
```diff
@@ -26,20 +30,116 @@
...
+        _ensure_directory_exists(dirname, mode=0o700)
+        data = AccessTokenStoreSchema().dump(self._store)
+        with open(self.cache_path, "w") as fp:
+            json.dump(data, fp, separators=(",", ":"))
+
+    def get(self, profile):
+        if self._store is None:
+            self._load_from_disk()
+        access_token = self._store.get(profile)
+        return access_token if _is_valid_access_token(access_token) else None
```
srstevenson:
```suggestion
        if _is_valid_access_token(access_token):
            return access_token
```
(although I'm also happy with explicitly returning `None` in the failure case).

faculty/session.py
srstevenson: "Asking forgiveness rather than permission" is often considered more Pythonic.

faculty/session.py
acroz: 👍

faculty/session.py
acroz: Yeah I have a preference for explicitly returning None in this case.

faculty/session.py
acroz: True, though we use `errno.ENOENT` elsewhere which is much less clear and I liked being consistent and having the comment in both places. What do you think?

srstevenson: Either way, consistency is good too.

zblz:
I wouldn't make this class extend `MemoryCache` as it is not conceptually an implementation of that.

Regarding breaking up the module into a subpackage, I think it makes sense. Having the caches alongside the Session class is a bit confusing, as when editing Session a dev only needs to know that some cache is used, not necessarily it's details.

PR #64, opened 2019-01-29
Add delete endpoint to server client

liamcoatman:


liamcoatman:
> Looks good, though should we return the raw response? If so, we should probably test the return value. My instinct is that the response is not valuable and we should just not return anything.

I've modified it to not return anything. 

> Also, does `_delete_raw` check the http status code?

Yes.

PR #65, opened 2019-01-29
Add endpoint to server client to get ssh details

liamcoatman:


srstevenson:
PEP 8 says:

> Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.

srstevenson:
PEP 8 says:
> Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.

zblz:
> PEP 8 says:
> 
> > Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.

I definitely agree, I find seeing `Http` or `Ssh` super weird.

PR #66, opened 2019-01-30
Add method to apply environment to server client

liamcoatman:


faculty/clients/server.py
```diff
@@ -186,3 +201,9 @@ def list(self, project_id, name=None):
         endpoint = "/instance/{}".format(project_id)
         params = {"name": name} if name is not None else None
         return self._get(endpoint, ServerSchema(many=True), params=params)
+
+    def apply_environment(self, environment_id, server_id):
+        endpoint = "/instance/{}/environment/{}".format(
+            server_id, environment_id
+        )
+        return self._put(endpoint, AppliedEnvironmentSchema())
```
zblz: I'm on the fence about the need to return anything here. At the end of the day, we are just giving back the arguments of the method after a convoluted roundtrip through Galleon, so I would be happy to remove the `return` keyword and delete the associated schema.

PR #73, opened 2019-02-08
Corrected error message in datasets _get_file.

mkunesch:
The source and destination paths in the error message in `datasets/__init__.py:_get_file` were the wrong way round. I have fixed this now.


acroz:
LGTM - will merge once tests have rerun.

PR #79, opened 2019-03-07
Add a method to list runs to the experiment client

tomas-milata:
Please check if you agree with the following semantics of the `experiment_ids` filter:

- `None` -> runs of all experiments returned
- `[]` -> no runs returned

Filtering by `lifecycle_stage` will be added in next versions.

Pagination is done in the same way as job runs. As discussed, we might want to factor it out later.

### DIFF
--- a/faculty/clients/experiment.py
+++ b/faculty/clients/experiment.py
@@ -103,7 +103,6 @@ class PageSchema(BaseSchema):
         return Page(**data)
 
 
-# TODO reuse pagination from jobs?
 class PaginationSchema(BaseSchema):
     start = fields.Integer(required=True)
     size = fields.Integer(required=True)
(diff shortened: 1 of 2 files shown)


faculty/clients/experiment.py
```diff
@@ -195,3 +230,65 @@ def get_run(self, project_id, run_id):
...
+        Get all experiment runs by making successive calls to ``list_runs``,
+        passing the ``start`` and ``limit`` of the ``next`` page each time
+        until ``next`` is returned as ``None``.
+
+        Parameters
+        ----------
+        project_id : uuid.UUID
+        experiment_ids : List[int], optional
+            To filter runs of experiments with the given IDs only. If an empty
+            list is passed, a result with an empty list of runs is returned.
```
pbugnion: Probably worth adding something like:
"By default, all experiments are returned"

tomas-milata: I fixed the commend.

tomas-milata: OK, let's factor out the pagination model in a separate PR.

PR #80, opened 2019-03-12
Log Run data

mociarain:
This PR adds models and schemas for Experiment Run Tags/Metrics/Params. An additional schema has been added called ExperimentRunDataSchema to wrap all these internally.

Note: To implement this an additional `_patch` & `_patch_raw` method were added to the base_client

### DIFF
--- a/faculty/clients/base.py
+++ b/faculty/clients/base.py
@@ -23,9 +23,10 @@ class InvalidResponse(Exception):
 
 
 class HttpError(Exception):
-    def __init__(self, response, error=None):
+    def __init__(self, response, error=None, error_code=None):
         self.response = response
         self.error = error
+        self.error_code = error_code
 
 
 HTTPError = HttpError  # For backwards compatiblity
@@ -92,6 +93,7 @@ class BaseSchema(Schema):
 
 class ErrorSchema(BaseSchema):
     error = fields.String(required=True)
+    error_code = fields.String(required=True)
 
 
 def _extract_error(response):
@@ -102,10 +104,20 @@ def _extract_error(response):
         return None
 
 
+def _extract_error_code(response):
+    try:
+        data = response.json()
+        return ErrorSchema().load(data)["errorCode"]
+    except (ValueError, ValidationError):
+        return None
+
+
 def _check_status(response):
     if response.status_code >= 400:
         cls = HTTP_ERRORS.get(response.status_code, HttpError)
-        raise cls(response, _extract_error(response))
+        raise cls(
+            response, _extract_error(response), _extract_error_code(response)
+        )
 
 
 def _deserialise_response(schema, response):
--- a/faculty/clients/base.py
+++ b/faculty/clients/base.py
@@ -171,6 +171,13 @@ class BaseClient(object):
         response = self._put_raw(endpoint, **kwargs)
         return _deserialise_response(schema, response)
 
+    def _patch_raw(self, endpoint, *args, **kwargs):
+        return self._request("PATCH", endpoint, *args, **kwargs)
+
+    def _patch(self, endpoint, schema, **kwargs):
+        response = self._patch_raw(endpoint, **kwargs)
+        return _deserialise_response(schema, response)
+
     def _delete_raw(self, endpoint, *args, **kwargs):
         return self._request("DELETE", endpoint, *args, **kwargs)
 
--- a/faculty/clients/base.py
+++ b/faculty/clients/base.py
@@ -103,7 +103,7 @@ def _check_status(response):
             data = ErrorSchema().load(response.json())
         except (ValueError, ValidationError):
             data = {}
-        raise cls(response, data.get("error"), data.get("errorCode"))
+        raise cls(response, data.get("error"), data.get("error_code"))
 
 
 def _deserialise_response(schema, response):
@@ -145,7 +145,6 @@ class BaseClient(object):
         url = self.session.service_url(self.SERVICE_NAME, endpoint)
         response = self.http_session.request(method, url, *args, **kwargs)
         if check_status:
-            print(response)
             _check_status(response)
         return response
 
--- a/faculty/clients/base.py
+++ b/faculty/clients/base.py
@@ -92,32 +92,18 @@ class BaseSchema(Schema):
 
 
 class ErrorSchema(BaseSchema):
-    error = fields.String(required=True)
-    error_code = fields.String(required=True)
-
-
-def _extract_error(response):
-    try:
-        data = response.json()
-        return ErrorSchema().load(data)["error"]
-    except (ValueError, ValidationError):
-        return None
-
-
-def _extract_error_code(response):
-    try:
-        data = response.json()
-        return ErrorSchema().load(data)["errorCode"]
-    except (ValueError, ValidationError):
-        return None
+    error = fields.String(missing=None)
+    error_code = fields.String(data_key="error_code", missing=None)
 
 
 def _check_status(response):
     if response.status_code >= 400:
         cls = HTTP_ERRORS.get(response.status_code, HttpError)
-        raise cls(
-            response, _extract_error(response), _extract_error_code(response)
-        )
+        try:
+            data = ErrorSchema().load(response.json())
+        except (ValueError, ValidationError):
+            data = {}
+        raise cls(response, data.get("error"), data.get("errorCode"))
 
 
 def _deserialise_response(schema, response):
@@ -159,6 +145,7 @@ class BaseClient(object):
         url = self.session.service_url(self.SERVICE_NAME, endpoint)
         response = self.http_session.request(method, url, *args, **kwargs)
         if check_status:
+            print(response)
             _check_status(response)
         return response
 
--- a/faculty/clients/base.py
+++ b/faculty/clients/base.py
@@ -93,7 +93,7 @@ class BaseSchema(Schema):
 
 class ErrorSchema(BaseSchema):
     error = fields.String(missing=None)
-    error_code = fields.String(data_key="error_code", missing=None)
+    error_code = fields.String(data_key="errorCode", missing=None)
 
 
 def _check_status(response):
--- a/faculty/clients/experiment.py
+++ b/faculty/clients/experiment.py
@@ -94,6 +94,22 @@ class ExperimentRunSchema(BaseSchema):
         return ExperimentRun(**data)
 
 
+class ExperimentRunTagSchema(BaseSchema):
+    key = fields.String(required=True)
+    value = fields.String(required=True)
+
+
+class ExperimentRunParameterSchema(BaseSchema):
+    key = fields.String(required=True)
+    value = fields.String(required=True)
+
+
+class ExperimentRunMetricSchema(BaseSchema):
+    key = fields.String(required=True)
+    value = fields.Float(required=True)
+    timestamp = fields.DateTime(missing=None)
+
+
 class PageSchema(BaseSchema):
     start = fields.Integer(required=True)
     limit = fields.Integer(required=True)
@@ -293,3 +309,62 @@ class ExperimentClient(BaseClient):
         return self._get(
             endpoint, ListExperimentRunsResponseSchema(), params=query_params
         )
+
+    def log_run_data(
+        self, project_id, run_id, tags=[], metrics=[], parameters=[]
+    ):
+        """Update the data of a run.
+
+        Parameters
+        ----------
+        project_id : uuid.UUID
+        run_id : uuid.UUID
+        tags : List[dict], optional
+            A list of tag value sets. Each set of tag values will result in an
+            upserted tag for each key passed on the specified run.
+            Example:
+            [
+                {
+                    'key': <str>,
+                    'value': <str>
+                },
+                {...}
+            ]
+        params : List[dict], optional
+            A list of parameter value sets. Each set of parameter values will
+            result in an upserted tag for each key passed on the specified run.
+            Example:
+            [                {
+                    'key': <str>,
+                    'value': <int>
+                },
+                {...}
+            ]
+        metrics : List[dict]
+            [
+                {
+                    'key': <str>,
+                    'value': <int>,
+                    'timestamp': datetime.datetime
+                },
+                {...}
+            ]}
+        Note: If the datetime does not have a
+            timezone, it will be assumed to be in UTC.
+
+        Returns
+        -------
+        None
+        """
+        endpoint = "/project/{}/run/{}/data".format(project_id, run_id)
+        payload = {
+            "tags": [ExperimentRunTagSchema().dump(tag) for tag in tags],
+            "params": [
+                ExperimentRunParameterSchema().dump(parameter)
+                for parameter in parameters
+            ],
+            "metrics": [
+                ExperimentRunMetricSchema().dump(metric) for metric in metrics
+            ],
+        }
+        return self._patch_raw(endpoint, json=payload)
--- a/faculty/clients/experiment.py
+++ b/faculty/clients/experiment.py
@@ -18,7 +18,16 @@ from enum import Enum
 from marshmallow import fields, post_load
 from marshmallow_enum import EnumField
 
-from faculty.clients.base import BaseSchema, BaseClient
+from faculty.clients.base import BaseClient, BaseSchema, Conflict
+
+class ParamConflictError(Exception):
+    def __init__(self, response, message, conflicting_params=None):
+        self.response = response
+        self.message = message
+        if conflicting_params is None:
+            self.conflicting_params = []
+        else:
+            self.conflicting_params = conflicting_params
 
 
 class ExperimentRunStatus(Enum):
@@ -357,4 +366,14 @@ class ExperimentClient(BaseClient):
         payload = ExperimentRunDataSchema().dump(
             {"metrics": metrics, "params": params, "tags": tags}
         )
-        return self._patch_raw(endpoint, json=payload)
+        try:
+            return self._patch_raw(endpoint, json=payload)
+        except Conflict as err:
+            if err.error_code == "conflicting_params":
+                raise ParamConflictError(
+                    err.response,
+                    err.error,
+                    err.response.json()["parameterKeys"],
+                )
+            else:
+                raise err
--- a/faculty/clients/experiment.py
+++ b/faculty/clients/experiment.py
@@ -353,7 +353,7 @@ class ExperimentClient(BaseClient):
         metrics : List[Metric], optional
             Each metric will be inserted.
         params : List[Param], optional
-            Each param will be inserted. Note that on a name conflict the 
+            Each param will be inserted. Note that on a name conflict the
             entire operation will be rejected.
         tags : List[Tag], optional
             Each tag be upserted.
--- a/faculty/clients/experiment.py
+++ b/faculty/clients/experiment.py
@@ -100,13 +100,14 @@ class ExperimentRunSchema(BaseSchema):
         return ExperimentRun(**data)
 
 
-class ExperimentRunTagSchema(BaseSchema):
+class ExperimentRunMetricSchema(BaseSchema):
     key = fields.String(required=True)
-    value = fields.String(required=True)
+    value = fields.Float(required=True)
+    timestamp = fields.DateTime(missing=None)
 
     @post_load
-    def make_experiment_run_tag(self, data):
-        return ExperimentRunTag(**data)
+    def make_experiment_run_metric(self, data):
+        return ExperimentRunMetric(**data)
 
 
 class ExperimentRunParamSchema(BaseSchema):
@@ -118,14 +119,13 @@ class ExperimentRunParamSchema(BaseSchema):
         return ExperimentRunParam(**data)
 
 
-class ExperimentRunMetricSchema(BaseSchema):
+class ExperimentRunTagSchema(BaseSchema):
     key = fields.String(required=True)
-    value = fields.Float(required=True)
-    timestamp = fields.DateTime(missing=None)
+    value = fields.String(required=True)
 
     @post_load
-    def make_experiment_run_metric(self, data):
-        return ExperimentRunMetric(**data)
+    def make_experiment_run_tag(self, data):
+        return ExperimentRunTag(**data)
 
 
 class ExperimentRunDataSchema(BaseSchema):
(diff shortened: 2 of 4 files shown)


faculty/clients/experiment.py
```diff
@@ -94,6 +100,40 @@ def make_experiment_run(self, data):
         return ExperimentRun(**data)
 
 
+class ExperimentRunTagSchema(BaseSchema):
```
acroz: nit: Can we reorder these the same as the classes above?

faculty/clients/experiment.py
```diff
@@ -293,3 +333,34 @@ def list_runs(
...
+    def log_run_data(
+        self, project_id, run_id, metrics=None, params=None, tags=None
+    ):
+        """Update the data of a run.
+
+        Parameters
+        ----------
+        project_id : uuid.UUID
+        run_id : uuid.UUID
+        metrics : List[experiments.ExperimentRunMetric], optional
```
acroz: I'd just document these as `List[ExperimentRunMetric]` etc. since `experiments.ExperimentRunMetric` doesn't give you the full import path anyway - better to just accept that these types are defined in the same context as the client.

faculty/clients/experiment.py
```diff
@@ -293,3 +333,34 @@ def list_runs(
...
+            Each tag be upserted.
+
+        Returns
+        -------
+        None
+        """
+        endpoint = "/project/{}/run/{}/data".format(project_id, run_id)
+        payload = ExperimentRunDataSchema().dump(
+            {
+                "metrics": ExperimentRunMetricSchema(many=True).dump(metrics),
```
acroz: The `ExperimentRunDataSchema` should use the `ExperimetnRunMetricSchema` etc. internally by itself - why have you pre-serialised the data here before passing it to `ExperimentRunDataSchema.dump` ?

faculty/clients/experiment.py
```diff
@@ -293,3 +333,34 @@ def list_runs(
...
+                "params": ExperimentRunParamSchema(many=True).dump(params),
+                "tags": ExperimentRunTagSchema(many=True).dump(tags),
+            }
+        )
+        return self._patch_raw(endpoint, json=payload)
```
acroz: I take it we're eventually going to interpret the response from the server to determine when e.g. metric updates have failed? Will this happen in a later PR?

tests/clients/test_experiment.py
```diff
@@ -92,6 +97,37 @@
...
+    key="other-tag-key", value="tag-value"
+)
+OTHER_EXPERIMENT_RUN_TAG_BODY = {"key": "other-tag-key", "value": "tag-value"}
+
+EXPERIMENT_RUN_PARAM = ExperimentRunParam(
+    key="parameter-key", value="tag-value"
+)
+EXPERIMENT_RUN_PARAM_BODY = {"key": "parameter-key", "value": "tag-value"}
+
+
```
acroz: nit: Remove one of the empty lines here?

tests/clients/test_experiment.py
```diff
@@ -138,11 +174,6 @@ def test_experiment_schema_invalid():
         ExperimentSchema().load({})
 
 
-def test_experiment_run_schema():
```
acroz: Why had this been removed?

faculty/clients/experiment.py
mociarain: It's only the params that can cause a 409 and I was going to handle that specifically in the plugin. I believe that it would be correctly mapped to a `faculty.clients.base.HttpError.Conflict` as it stands now?

tests/clients/test_experiment.py
```diff
@@ -363,3 +409,65 @@ def test_experiment_client_list_runs_page(mocker):
         schema_mock.return_value,
         params=[("start", 20), ("limit", 10)],
     )
+
+
+def test_log_metadata(mocker):
```
acroz: `test_log_metadata` -> `test_log_run_data` (also tests below)?

faculty/clients/experiment.py
acroz: > I was going to handle that specifically in the plugin.

That might be the right solution, however since it's usually the responsibility of a client to deserialise information in the response we might want to consider converting the 409 to a custom exception type with structured information inside in the client.

> I believe that it would be correctly mapped to a `faculty.clients.base.HttpError.Conflict` as it stands now?

Yeah, I think so.

faculty/clients/experiment.py
```diff
@@ -55,6 +55,12 @@ class ExperimentRunStatus(Enum):
     ],
 )
 
+ExperimentRunMetric = namedtuple(
+    "ExperimentRunMetric", ["key", "value", "timestamp"]
+)
+ExperimentRunParam = namedtuple("ExperimentRunParam", ["key", "value"])
+ExperimentRunTag = namedtuple("ExperimentRunTag", ["key", "value"])
```
pbugnion: Should these just be called `Metric`, `Param` and `Tag`?

faculty/clients/experiment.py
pbugnion: The server can also return a 404 if the run does not exist.

pbugnion: Also, the server gives back a detailed message in case there is a parameter conflict. It feels like we can do a lot more here, as Andrew said.

pbugnion:
Open for discussion, but is there value in not dispatching the request if all the fields are empty?

faculty/clients/experiment.py
mociarain: Done

faculty/clients/experiment.py
mociarain: 👍

faculty/clients/experiment.py
mociarain: I thought I needed to. Changed

tests/clients/test_experiment.py
mociarain: 👍

tests/clients/test_experiment.py
mociarain: 👍

faculty/clients/experiment.py
mociarain: I think that's a good call

faculty/clients/experiment.py
```diff
@@ -293,3 +331,30 @@ def list_runs(
...
+        ----------
+        project_id : uuid.UUID
+        run_id : uuid.UUID
+        metrics : List[Metric], optional
+            Each metric will be inserted.
+        params : List[Metric], optional
+            Each param will be inserted and rejected if the given key is
+            present.
+        tags : List[Tag], optional
+            Each tag be upserted.
```
pbugnion: Should we add a note to say -- if any params conflict, the entire operation is rejected?

faculty/clients/experiment.py
mociarain: After chatting with Pascal we decided to implement a custom error for the Conflict case but not for the 404 as we don't in the other methods. This doesn't mean that we shouldn't do it and retrofit it to existing methods but it won't be part of this PR

faculty/clients/experiment.py
mociarain: 👍

tests/clients/test_base.py
```diff
@@ -16,24 +16,23 @@
 from collections import namedtuple
```
acroz: Should we also add a test for the `ErrorSchema` as discussed?

tests/clients/test_experiment.py
```diff
@@ -92,6 +99,30 @@
     "endedAt": RUN_ENDED_AT_STRING,
     "deletedAt": DELETED_AT_STRING,
 }
+TAG = Tag(key="tag-key", value="tag-value")
+TAG_BODY = {"key": "tag-key", "value": "tag-value"}
+
+OTHER_TAG = Tag(key="other-tag-key", value="tag-value")
+OTHER_TAG_BODY = {"key": "other-tag-key", "value": "tag-value"}
+
+PARAM = Param(key="parameter-key", value="tag-value")
```
acroz: `param-value`?

acroz: (and line below)

tests/clients/test_experiment.py
```diff
@@ -92,6 +99,30 @@
     "endedAt": RUN_ENDED_AT_STRING,
     "deletedAt": DELETED_AT_STRING,
 }
+TAG = Tag(key="tag-key", value="tag-value")
+TAG_BODY = {"key": "tag-key", "value": "tag-value"}
+
+OTHER_TAG = Tag(key="other-tag-key", value="tag-value")
```
acroz: `other-tag-value`?

acroz: (and line below)

faculty/clients/base.py
```diff
@@ -91,21 +92,18 @@ class Meta:
...
 
 def _check_status(response):
     if response.status_code >= 400:
         cls = HTTP_ERRORS.get(response.status_code, HttpError)
-        raise cls(response, _extract_error(response))
+        try:
+            data = ErrorSchema().load(response.json())
+        except (ValueError, ValidationError):
+            data = {}
+        raise cls(response, data.get("error"), data.get("errorCode"))
```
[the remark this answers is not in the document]
mociarain: Yes

faculty/clients/experiment.py
```diff
@@ -18,7 +18,16 @@
 from marshmallow import fields, post_load
 from marshmallow_enum import EnumField
 
-from faculty.clients.base import BaseSchema, BaseClient
+from faculty.clients.base import BaseClient, BaseSchema, Conflict
+
+
+class ParamConflict(Exception):
+    def __init__(self, message, conflicting_params=None):
+        self.message = message
```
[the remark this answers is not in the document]
mociarain: You're right. Sorry I missed that

faculty/clients/experiment.py
```diff
@@ -94,6 +107,40 @@ def make_experiment_run(self, data):
         return ExperimentRun(**data)
 
 
+class MetricSchema(BaseSchema):
+    key = fields.String(required=True)
+    value = fields.Float(required=True)
+    timestamp = fields.DateTime(missing=None)
```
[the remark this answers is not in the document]
mociarain: It should never be. I'll change that

tests/clients/test_base.py
```diff
@@ -16,24 +16,23 @@
...
     InternalServerError,
-    BadGateway,
+    InvalidResponse,
+    MethodNotAllowed,
+    NotFound,
     ServiceUnavailable,
-    GatewayTimeout,
+    Unauthorized,
 )
+from marshmallow import fields, post_load
```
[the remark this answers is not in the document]
mociarain: 👍

tests/clients/test_base.py
```diff
@@ -16,24 +16,23 @@
 from collections import namedtuple
 
 import pytest
-from marshmallow import fields, post_load
-
 from faculty.clients.base import (
```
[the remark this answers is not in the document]
mociarain: I was ordering the imports from faculty.clients.base and got a little carried away.

tests/clients/test_base.py
mociarain: Do you think it's worth it? It's 2 optional strings

tests/clients/test_experiment.py
mociarain: 👍

tests/clients/test_experiment.py
mociarain: 👍

tests/clients/test_base.py
mociarain: 👍

faculty/clients/experiment.py
```diff
@@ -293,3 +340,40 @@ def list_runs(
...
+            Each param will be inserted. Note that on a name conflict the
+            entire operation will be rejected.
+        tags : List[Tag], optional
+            Each tag be upserted.
+
+        Returns
+        -------
+        None
+        """
+        if all([t is None for t in [metrics, params, tags]]):
```
acroz: Maybe `kwarg` would be a more descriptive name instead of `t`?

faculty/clients/experiment.py
```diff
@@ -293,3 +340,40 @@ def list_runs(
...
+            Each metric will be inserted.
+        params : List[Param], optional
+            Each param will be inserted. Note that on a name conflict the
+            entire operation will be rejected.
+        tags : List[Tag], optional
+            Each tag be upserted.
+
+        Returns
+        -------
+        None
```
acroz: I think that typically in this case you'd just omit the `Returns` block completely since we're kind of not returning anything.

tests/clients/test_base.py
```diff
@@ -56,7 +57,14 @@
     (418, HttpError),
 ]
 
-HTTP_METHODS = ["GET", "POST", "PUT", "DELETE"]
+HTTP_METHODS = ["GET", "POST", "PUT", "DELETE", "PATCH"]
+
+
+def test_error_schema():
```
acroz: 👍

tests/clients/test_experiment.py
```diff
@@ -92,6 +99,30 @@
     "endedAt": RUN_ENDED_AT_STRING,
     "deletedAt": DELETED_AT_STRING,
 }
+TAG = Tag(key="tag-key", value="tag-value")
+TAG_BODY = {"key": "tag-key", "value": "tag-value"}
+
+OTHER_TAG = Tag(key="other-tag-key", value="other-tag-value")
+OTHER_TAG_BODY = {"key": "other-tag-key", "value": "other-tag-value"}
+
+PARAM = Param(key="param-key", value="tag-value")
```
acroz: `"param-value"` (and on the line below)?

faculty/clients/experiment.py
mociarain: 👍

tests/clients/test_experiment.py
mociarain: 👍

mociarain: 👍

tests/clients/test_experiment.py
```diff
@@ -169,6 +195,41 @@ def test_create_run_schema(started_at, artifact_location):
...
+    data = ParamSchema().load(PARAM_BODY)
+    assert data == PARAM
+
+
+def test_experiment_run_tag_schema():
+    data = TagSchema().load(TAG_BODY)
+    assert data == TAG
+
+
+def test_experiment_run_data_schema():
```
acroz: We're actually using the `ExperimentRunDataSchema` to `dump()`, not `load()`. Can we invert these tests to test that?

faculty/clients/experiment.py
mociarain: 👍

faculty/clients/experiment.py
mociarain: 👍

tests/clients/test_experiment.py
```diff
@@ -363,3 +424,80 @@ def test_experiment_client_list_runs_page(mocker):
...
+
+    client = ExperimentClient(mocker.Mock())
+    client.log_run_data(
+        PROJECT_ID,
+        EXPERIMENT_RUN_ID,
+        metrics=[METRIC],
+        params=[PARAM],
+        tags=[TAG],
+    )
+    expected_body = {"metrics": [METRIC], "params": [PARAM], "tags": [TAG]}
```
acroz: I found the name here (`expected_body`) a little confusing as I anticipated that we'd be asserting that this was the actualy dispatched JSON, whereas we're actually asserting that this is the thing passed to the schema's dump method. Can we consider renaming to something like `expected_dump_arg` or just inline it into the `assert_called_once_with` below?

tests/clients/test_experiment.py
```diff
@@ -363,3 +424,80 @@ def test_experiment_client_list_runs_page(mocker):
...
+    run_data_schema_mock.assert_called_once_with()
+    run_data_dump_mock.assert_called_once_with(expected_body)
+    ExperimentClient._patch_raw.assert_called_once_with(
+        "/project/{}/run/{}/data".format(PROJECT_ID, EXPERIMENT_RUN_ID),
+        json=run_data_dump_mock.return_value,
+    )
+
+
+def test_log_run_data_param_conflict(mocker):
+    expected_message = "bad params"
```
acroz: Since these are fixtures rather than expected return values from something, can we remove the `expected_` prefix?

tests/clients/test_experiment.py
```diff
@@ -363,3 +424,80 @@ def test_experiment_client_list_runs_page(mocker):
...
+    expected_error_code = "conflicting_params"
+    json_mock = mocker.Mock()
+    json_mock.json.return_value = {"parameterKeys": ["bad-key"]}
+    response_mock = Conflict(json_mock, expected_message, expected_error_code)
```
acroz:
```suggestion
    exception = Conflict(json_mock, expected_message, expected_error_code)
```
Since it's not a `Response` object and it's an actual exception rather than a mock object?

tests/clients/test_experiment.py
```diff
@@ -363,3 +424,80 @@ def test_experiment_client_list_runs_page(mocker):
...
+    )
+    run_data_schema_mock = mocker.patch(
+        "faculty.clients.experiment.ExperimentRunDataSchema"
+    )
+
+    client = ExperimentClient(mocker.Mock())
+
+    with pytest.raises(ParamConflict, match=expected_message):
+        client.log_run_data(PROJECT_ID, EXPERIMENT_RUN_ID, params=[PARAM])
+    run_data_schema_mock.assert_called_once_with()
```
acroz: Could probably remove this assertion since we're not checking if the schema's `dump` method call either. This would mean we don't need to bind the `run_data_schema_mock` above either.

tests/clients/test_experiment.py
```diff
@@ -363,3 +424,80 @@ def test_experiment_client_list_runs_page(mocker):
...
+
+
+def test_log_run_data_other_conflict(mocker):
+    json_mock = mocker.Mock()
+    response_mock = Conflict(json_mock, "", "")
```
acroz:
```suggestion
    exception = Conflict(json_mock, "", "")
```
as above.

tests/clients/test_experiment.py
mociarain: 👍

tests/clients/test_experiment.py
```diff
@@ -363,3 +424,80 @@ def test_experiment_client_list_runs_page(mocker):
...
+    )
+    run_data_schema_mock = mocker.patch(
+        "faculty.clients.experiment.ExperimentRunDataSchema"
+    )
+
+    client = ExperimentClient(mocker.Mock())
+
+    with pytest.raises(Conflict):
+        client.log_run_data(PROJECT_ID, EXPERIMENT_RUN_ID, params=[PARAM])
+    run_data_schema_mock.assert_called_once_with()
```
acroz: Remove this line and don't bind `run_data_schema_mock` as above?

tests/clients/test_experiment.py
mociarain: 👍

faculty/clients/experiment.py
acroz: 👍

tests/clients/test_experiment.py
mociarain: 👍

tests/clients/test_experiment.py
mociarain: 👍

tests/clients/test_experiment.py
mociarain: 👍

tests/clients/test_experiment.py
mociarain: 👍

tests/clients/test_experiment.py
mociarain: 👍

tests/clients/test_experiment.py
```diff
@@ -169,6 +196,43 @@ def test_create_run_schema(started_at, artifact_location):
...
+
+def test_experiment_run_data_schema():
+    data = ExperimentRunDataSchema().dump(
+        {"metrics": [METRIC], "params": [PARAM], "tags": [TAG]}
+    )
+    assert data == EXPERIMENT_RUN_DATA_BODY
+
+
+def test_experiment_run_data_schema_empty():
+    data = ExperimentRunDataSchema().load({})
```
acroz: As above, this should test `dump()`

tests/clients/test_experiment.py
```diff
@@ -169,6 +196,43 @@ def test_create_run_schema(started_at, artifact_location):
...
+    assert data == {}
+
+
+def test_experiment_run_data_schema_multiple():
+    data = ExperimentRunDataSchema().load({"tags": [TAG_BODY, OTHER_TAG_BODY]})
```
acroz: Ditto

tests/clients/test_experiment.py
acroz: You marked this as resolved but the change does not appear to have been applied to this test (though it has been applied to the later ones)

tests/clients/test_experiment.py
```diff
@@ -363,3 +427,68 @@ def test_experiment_client_list_runs_page(mocker):
...
+    ExperimentClient._patch_raw.assert_called_once_with(
+        "/project/{}/run/{}/data".format(PROJECT_ID, EXPERIMENT_RUN_ID),
+        json=run_data_dump_mock.return_value,
+    )
+
+
+def test_log_run_data_param_conflict(mocker):
+    message = "bad params"
+    error_code = "conflicting_params"
+    json_mock = mocker.Mock()
```
acroz: I wouldn't have bothered commenting on this if there weren't other changes I'd commented on, but `json_mock` would probably better be `response_mock` in this case, since it's really a mock `requests.Response` object.

tests/clients/test_experiment.py
mociarain: 👍

tests/clients/test_experiment.py
mociarain: 👍

tests/clients/test_experiment.py
mociarain: 👍

tests/clients/test_experiment.py
mociarain: Done now

tomas-milata:
I we'd normally merge this when the backend is deployed. But it will be shortly there.

PR #81, opened 2019-03-13
Pass empty tags on run creation, return tags

tomas-milata:
At the moment, the backend requires `tags` to be passed in the request to create a run. This fixes it by passing empty tags for now.

The next step is to pass through tags from mlflow-faculty to this client.

Also, tag schema was added to the run schema (used when fetching runs).

### DIFF
--- a/faculty/clients/experiment.py
+++ b/faculty/clients/experiment.py
@@ -62,6 +62,8 @@ ExperimentRun = namedtuple(
         "ended_at",
         "deleted_at",
         "tags",
+        "params",
+        "metrics"
     ],
 )
 
@@ -131,6 +133,8 @@ class ExperimentRunSchema(BaseSchema):
     ended_at = fields.DateTime(data_key="endedAt", missing=None)
     deleted_at = fields.DateTime(data_key="deletedAt", missing=None)
     tags = fields.Nested(TagSchema, many=True, required=True)
+    params = fields.Nested(ParamSchema, many=True, required=True)
+    metrics = fields.Nested(MetricSchema, many=True, required=True)
 
     @post_load
     def make_experiment_run(self, data):
--- a/faculty/clients/experiment.py
+++ b/faculty/clients/experiment.py
@@ -202,7 +202,12 @@ class ExperimentClient(BaseClient):
         return self._get(endpoint, ExperimentSchema(many=True))
 
     def create_run(
-        self, project_id, experiment_id, started_at, artifact_location=None
+        self,
+        project_id,
+        experiment_id,
+        started_at,
+        artifact_location=None,
+        tags=[],
     ):
         """Create a run in a project.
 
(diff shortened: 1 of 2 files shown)


faculty/clients/experiment.py
```diff
@@ -203,6 +222,7 @@ def create_run(
             The location of the artifact repository to use for this run.
             If omitted, the value of `artifact_location` for the experiment
             will be used.
+        tags: dict[str, str]
```
acroz: Shouldn't this be `List[Tag]`?

tests/clients/test_experiment.py
```diff
@@ -121,6 +128,11 @@
 }
 
 
+def test_tag_schema():
```
acroz: We should probably test the `.dump()` on this as well as it's being used for serialisation.

tests/clients/test_experiment.py
```diff
@@ -121,6 +128,11 @@
...
+    data = TagSchema().load(TAG_BODY)
+    assert data == TAG
+
+
 def test_experiment_schema():
```
acroz: Can we also test the case where non-empty tags are passed?

faculty/clients/experiment.py
```diff
@@ -188,7 +202,12 @@ def list(self, project_id):
         return self._get(endpoint, ExperimentSchema(many=True))
 
     def create_run(
-        self, project_id, experiment_id, started_at, artifact_location=None
+        self,
+        project_id,
+        experiment_id,
+        started_at,
+        artifact_location=None,
+        tags=[],
```
acroz: Having mutable objects as default arguments is discouraged in Python for reasons I'd rather not type out (can explain in person). Recommend having `tags=None` here and then in the method have:

```python
if tags is None:
    tags = []
```

faculty/clients/experiment.py
tomas-milata: yes, thanks

tests/clients/test_experiment.py
tomas-milata: I think this is now covered by the tests that Maitiu added and the one that I added.

tests/clients/test_experiment.py
```diff
@@ -216,6 +229,11 @@ def test_experiment_run_tag_schema():
     assert data == TAG
 
 
+def test_experiment_run_tag_schema_dump():
```
acroz: This should probably just be `test_tag_schema_dump`.

Usually test names are in the format `test_foo_bar_method_name` when testing `FooBar.method_name`. If other tests break this we should fix. (Exceptions are usually made for schemas only used to convert in one direction, as is the case in most client modules in this library).

tomas-milata: OK, I renamed it, along with other similar tests.

PR #82, opened 2019-03-14
Add method to update experiment metadata

acroz:


acroz:
> If name and description are unset we could not call the client. This is done in the `log_run_data`method

Good call - I did not handle this case in the server. I've opened a PR to correct this on the server side.

acroz:
This is ready for re-review.

I need to make updates to the corresponding PR on mlflow-faculty.

tomas-milata:
@acroz What would you say that is our convention for raising custom exceptions based on 4XX statuses? Should we always do it? Or only when we need to return some specific data from the error response to the user? I'm not completely sure and want to stick to some convention in other PRs.

acroz:
> @acroz What would you say that is our convention for raising custom exceptions based on 4XX statuses? Should we always do it? Or only when we need to return some specific data from the error response to the user? I'm not completely sure and want to stick to some convention in other PRs.

@tomas-milata I think we should use them when they represent aspects of program flows that we'd like to handle. Here, the caller may not know ahead of time that they're setting an experiment name that's taken, and so being able to catch explicitly an `ExperimentNameConflict` exception seems valuable.

I know there's a soft distinction between this and `NotFound` cases, which we arguably might want to handle in a similar way, but more practically we probably want to avoid a proliferation of custom exception types for every resource in the library.

Maybe a good test is 'would the caller of the client want to check the 'errorCode' on the response if we didn't convert to our own exception type'? I think the caller should never have to do introspection of the content of the HTTP response as part of normal program flow.

tomas-milata:
> > @acroz What would you say that is our convention for raising custom exceptions based on 4XX statuses? Should we always do it? Or only when we need to return some specific data from the error response to the user? I'm not completely sure and want to stick to some convention in other PRs.
> 
> @tomas-milata I think we should use them when they represent aspects of program flows that we'd like to handle. Here, the caller may not know ahead of time that they're setting an experiment name that's taken, and so being able to catch explicitly an `ExperimentNameConflict` exception seems valuable.
> 
> I know there's a soft distinction between this and `NotFound` cases, which we arguably might want to handle in a similar way, but more practically we probably want to avoid a proliferation of custom exception types for every resource in the library.
> 
> Maybe a good test is 'would the caller of the client want to check the 'errorCode' on the response if we didn't convert to our own exception type'? I think the caller should never have to do introspection of the content of the HTTP response as part of normal program flow.

OK, this makes sense, thank you!

tomas-milata:
Sorry, I accidentally closed it.

PR #83, opened 2019-03-18
Allow filtering experiment listing by lifecycle stage

eliasbenussi:


faculty/clients/experiment.py
```diff
@@ -220,19 +225,28 @@ def get(self, project_id, experiment_id):
...
+            (ACTIVE_ONLY | DELETED_ONLY). By default, all experiments in the
+            project are returned.
 
         Returns
         -------
         List[Experiment]
         """
+        query_params = []
+        if lifecycle_stage is not None:
+            query_params.append(("lifecycleStage", lifecycle_stage.value))
```
acroz:
```suggestion
            query_params["lifecycleStage"] = lifecycle_stage.value
```

faculty/clients/experiment.py
```diff
@@ -75,6 +75,11 @@ class ExperimentRunStatus(Enum):
 )
 
 
+class LifecycleStageFilter(Enum):
```
acroz: Since the input name to the method you're adding and the query parameter are 'lifecycle stage', maybe this should be to? And then do:

```python
class LifecycleStage(Enum):
    ACTIVE = "active"
    DELETED = "deleted"
```

faculty/clients/experiment.py
acroz: Oh wait, it's actually a list of tuples.

acroz: You can actually pass the query params as a dict, which is what I was thinking:
http://docs.python-requests.org/en/master/user/quickstart/#passing-parameters-in-urls

That seems to be the default way `requests` encourages you to do it but if you strongly prefer to keep a list of tuples that's fine.

eliasbenussi: I don't really know what is more conventional, I am happy with either. The only reason I did it like that is because that is what we do elsewhere in the code.
Whatever we decide I think we should have internal consistency

faculty/clients/experiment.py
eliasbenussi: Agreed

faculty/clients/experiment.py
acroz: I think we have lists of tuples in this module, but dict elsewhere in the library, though I think some of the other use cases in this module require specifying multiple experiment IDs, which probable requires the 'list of tuples' solution.

pbugnion: `query_params = {'lifecycleStage': lifecycle_stage.value} if lifecycle_stage is not None else {}`?

acroz: @pbugnion given that that probably won't fit on a single line I think not using the ternary operator will be clearer.

tomas-milata: It seems cleaner to me to use a `dict`. And use a list of tuples only when you need duplicate keys.

PR #84, opened 2019-03-18
Add a client method to delete a single experiment

tomas-milata:


faculty/clients/experiment.py
```diff
@@ -241,6 +241,24 @@ def list(self, project_id):
         endpoint = "/project/{}/experiment".format(project_id)
         return self._get(endpoint, ExperimentSchema(many=True))
 
+    def delete(self, project_id, experiment_id):
+        """Delete a specified experiment.
+
+        Parameters
+        ----------
+
```
acroz: Would remove this line.

faculty/clients/experiment.py
```diff
@@ -241,6 +241,24 @@ def list(self, project_id):
...
+        project_id : uuid.UUID
+        experiment_id : int
+
+        Returns
```
acroz: Typically, if a method doesn't return anything (== returns None), we'd just omit the 'Returns' block.

tests/clients/test_experiment.py
```diff
@@ -510,3 +510,15 @@ def test_log_run_data_empty(mocker):
...
     ExperimentClient._patch_raw.assert_not_called()
+
+
+def test_delete(mocker):
+    mocker.patch.object(ExperimentClient, "_delete_raw")
+
+    client = ExperimentClient(mocker.Mock())
+    result = client.delete(PROJECT_ID, EXPERIMENT_ID)
+
+    assert result is None
```
acroz: Similar to my comment on documentation, we'd not typically assert that the return value of a method in this case.

faculty/clients/experiment.py
tomas-milata: okay

PR #87, opened 2019-03-21
Add run name and parent run id as attributes

acroz:


acroz:
I've tested this against the associated Atlas PR.

pbugnion:
This looks good. Merging this will break the tests in mlflow-faculty. Is it worth putting in a PR to fix those?

PR #89, opened 2019-03-21
Implement and test update_run_info

eliasbenussi:


faculty/clients/experiment.py
```diff
@@ -479,3 +484,22 @@ def log_run_data(
...
+        """Update the status and end time of a run.
+
+        Parameters
+        ----------
+        project_id : uuid.UUID
+        run_id : uuid.UUID
+        status: ExperimentRunStatus, optional
+        ended_at: datetime, optional
+
+        """
```
pbugnion: We normally have a `return` section in the docstring.

faculty/clients/experiment.py
```diff
@@ -479,3 +484,22 @@ def log_run_data(
...
+        if status is None and ended_at is None:
+            return
```
pbugnion: I don't think that can work. We need to return the run anyway?

faculty/clients/experiment.py
```diff
@@ -479,3 +484,22 @@ def log_run_data(
...
+        endpoint = "/project/{}/run/{}/info".format(project_id, run_id)
+        payload = ExperimentRunInfoSchema().dump(
+            {"status": status, "ended_at": ended_at}
+        )
+        self._patch_raw(endpoint, json=payload)
```
pbugnion: We should be deserializing the run out of this.

PR #90, opened 2019-03-22
Install marshmallow recommended dependencies

acroz:
This is to correct an error where python-dateutil is installed, but of a version incompatible with marshmallow. The [reco] extras ensures that a version supported by marshmallow is installed.


pbugnion:
Agreed. The reason is we need marshmallow with datetime support.

PR #91, opened 2019-03-25
Add a method to restore an experiment to the client

tomas-milata:


acroz:
Just need to fix a flake8 linting error in setup.py AFAICT.

PR #92, opened 2019-03-25
Read region from secret service response

mociarain:
[DNM until https://bitbucket.org/theasi/secret-service/pull-requests/112/send-region-information/diff is merged]
This PR changes the datasets feature to use the AWS region returned from the secret-service as opposed to a hardcoded value.


mociarain:
This has been tested out on a dev-cluster in eu-west-1

PR #95, opened 2019-04-03
Removed specific encryption format

haileyfong:
Originally, client was mandating encryption AES256. We thought it is a better idea to use the default encryption format on the server so we removed all the specific encryption formats on the client side.

We tested it on fantomas, one of our test clusters, by checking if the put, copy and move commands that originally included the AES256 encryption breaks or not, then we checked if the newly put/copied/moved items are of the server default encryption format. 

Left to do:
 - [x] Check the above also works on the existing cluster.
 - [x] Format using black


acroz:
Looks like we just need to run black on this.

pbugnion:
Note that this is fully backwards compatible with existing deployments, provided the bucket encryption settings are correct (I checked that they were).

PR #96, opened 2019-04-16
Handle experiment not active error

eliasbenussi:


eliasbenussi:
In this PR we handle run creation receiving a 409 response due to the experiment being not active

tests/clients/test_experiment.py
```diff
@@ -503,6 +504,30 @@ def test_experiment_create_run(mocker):
     )
 
 
+def test_experiment_create_run_experiment_not_active_conflict(mocker):
+    message = "experiment not active"
+    error_code = "experiment_not_active"
+    response_mock = mocker.Mock()
+    response_mock.json.return_value = {"experimentId": "test-id"}
```
tomas-milata: It seems to me we don't need to mock the json return value, because we're relying on the base client to parse the JSON into the `Conflict` object.

tests/clients/test_experiment.py
```diff
@@ -503,6 +504,30 @@ def test_experiment_create_run(mocker):
...
+    exception = Conflict(response_mock, message, error_code)
+
+    mocker.patch.object(ExperimentClient, "_post", side_effect=exception)
+    started_at = mocker.Mock()
+    artifact_location = mocker.Mock()
+
+    client = ExperimentClient(mocker.Mock())
+    with pytest.raises(ExperimentNotActiveConflict, match=message):
+        # client.log_run_data(PROJECT_ID, EXPERIMENT_RUN_ID, params=[PARAM])
```
tomas-milata: You probably forgot to remove this.

tests/clients/test_experiment.py
tomas-milata: Sorry, I just noticed we're extracting the experiment ID... Ignore me.

faculty/clients/experiment.py
```diff
@@ -37,6 +37,12 @@ def __init__(self, message, conflicting_params=None):
             self.conflicting_params = conflicting_params
 
 
+class ExperimentNotActiveConflict(Exception):
```
acroz: What about `ExperimentDeleted`?

The fact that it's a conflict doesn't seem relevant (that seems to mainly be a feature of the HTTP status code we're using) and 'Deleted' seems clearer (and closer to our data model) than 'NotActive'/

faculty/clients/experiment.py
```diff
@@ -416,6 +422,12 @@ def create_run(
         Returns
         -------
         ExperimentRun
+
+        Raises
+        ------
+        ExperimentNotActiveConflict
```
acroz: Note to update this if we change the class name.

faculty/clients/experiment.py
```diff
@@ -416,6 +422,12 @@ def create_run(
...
+            When the run that is being updated refers to an experiment that is
+            deleted or does not exist
```
acroz: I would have thought we would return a 404 from the API in the case the experiment does not exist?

faculty/clients/experiment.py
```diff
@@ -548,6 +572,10 @@ def log_run_data(
                 raise ParamConflict(
                     err.error, err.response.json()["parameterKeys"]
                 )
+            elif err.error_code == "experiment_not_active":
```
acroz: Aren't we getting rid of this case?

faculty/clients/experiment.py
eliasbenussi: We do. I am not quite sure I understand how we deal with the NOT_FOUND cases (which are also not caught in other endpoints). It seems to me like we just return the default `NotFound` exception from the `base` client. Do you think this is not correct?

faculty/clients/experiment.py
```diff
@@ -416,6 +422,12 @@ def create_run(
         Returns
         -------
         ExperimentRun
+
+        Raises
+        ------
+        ExperimentDeleted
+            When the run that is being updated refers to an
```
acroz: Is this line wrapped unusually early?

faculty/clients/experiment.py
```diff
@@ -432,7 +444,19 @@ def create_run(
                 "tags": tags,
             }
         )
-        return self._post(endpoint, ExperimentRunSchema(), json=payload)
+        try:
+            return self._post(endpoint, ExperimentRunSchema(), json=payload)
+        except Conflict as err:
+            if err.error_code == "conflicting_params":
```
acroz: I don't think this endpoint ever returns this code.

tests/clients/test_experiment.py
```diff
@@ -503,6 +504,29 @@ def test_experiment_create_run(mocker):
     )
 
 
+def test_experiment_create_run_experiment_deleted_conflict(mocker):
+    message = "experiment deleted"
+    error_code = "experiment_deleted"
+    response_mock = mocker.Mock()
+    response_mock.json.return_value = {"experimentId": "test-id"}
```
acroz: Experiment ID is usually an `int` - maybe we should use an int as the mock value here?

tests/clients/test_experiment.py
```diff
@@ -503,6 +504,29 @@ def test_experiment_create_run(mocker):
...
+    exception = Conflict(response_mock, message, error_code)
+
+    mocker.patch.object(ExperimentClient, "_post", side_effect=exception)
+    started_at = mocker.Mock()
+    artifact_location = mocker.Mock()
+
+    client = ExperimentClient(mocker.Mock())
+    with pytest.raises(ExperimentDeleted, match=message):
```
acroz: Inlining `message` here might improve readability.

PR #101, opened 2019-04-19
Enhance environment client

ameroueh:
The environment client currently only returns high level
metadata about an environment. It also only allows one to
list environments from a project; more advanced operations
are not possible. This PR enhances this client
in several ways:

* The "specification" field is added to EnvironmentSchema

* A number of schemas (and associated objects) are introduced
to serialise/deserialise the "specification" field

* New methods are added to EnvironmentClient (get, create,
update, delete)

* Additional schemas (and associated objects) are
introduced to support the new methods of EnvironmentClient
(EnvironmentUpdateSchema, EnvironmentCreationResponseSchema)


faculty/clients/environment.py
```diff
@@ -42,16 +186,77 @@ class EnvironmentSchema(BaseSchema):
...
+        endpoint = "/project/{}/environment/{}".format(
+            project_id, environment_id
+        )
+        return self._get(endpoint, EnvironmentSchema())
+
+    def update(self, project_id, environment_id, update):
+        endpoint = "/project/{}/environment/{}".format(
+            project_id, environment_id
+        )
+        return self._put(
```
pbugnion: If we don't care about the return, I think you can just use `self._put_raw`. It then doesn't try to do anything with the return value.

faculty/clients/environment.py
```diff
@@ -42,16 +186,77 @@ class EnvironmentSchema(BaseSchema):
...
+            endpoint,
+            EnvironmentCreationResponseSchema(),
+            json=EnvironmentUpdateSchema().dump(content),
+        )
+
+    def delete(self, project_id, environment_id):
+        endpoint = "/project/{}/environment/{}".format(
+            project_id, environment_id
+        )
+        return self._delete(endpoint, schema=DummySchema())
```
pbugnion: Same as above.

faculty/clients/environment.py
```diff
@@ -42,16 +186,77 @@ class EnvironmentSchema(BaseSchema):
...
+    project_id = fields.UUID(data_key="projectId", required=True)
+    name = fields.String(required=True)
+    description = fields.String(required=True)
+    author_id = fields.UUID(data_key="authorId", required=True)
+    specification = fields.Nested(SpecificationSchema(), required=True)
 
     @post_load
     def make_environment(self, data):
+        data["created_at"] = None
+        data["updated_at"] = None
```
pbugnion: Returning an environment that's almost fully formed but not quite is a bit awkward.

I think there's an argument for only returning the ID.

pbugnion:
This is great! Thanks a lot.

I made a couple of comments inline.

Did you check the integration against a deployment?

ameroueh:
> Did you check the integration against a deployment?

I have a script which uses each method of the client successfully, but I do not have proper integration tests.

faculty/clients/environment.py
ameroueh: I completely agree - I assume that it's easier for me to just return the environment ID than for you to modify the environment creation endpoint, so I'll do that. (Though I do think these two fields should be returned by the environment creation endpoint if possible, for consistency.)

pbugnion:
> I have a script which uses each method of the client successfully, but I do not have proper integration tests.

That's fine -- I was just making sure this had actually been checked against real payloads.

faculty/clients/environment.py
pbugnion: @acroz @zblz what do you think?

faculty/clients/environment.py
```diff
@@ -42,16 +186,77 @@ class EnvironmentSchema(BaseSchema):
...
+    def make_environment(self, data):
+        return Environment(**data)
+
+
+class EnvironmentUpdateSchema(BaseSchema):
+    name = fields.String(required=True)
+    description = fields.String(required=True)
+    specification = fields.Nested(SpecificationSchema(), required=True)
+
+    @post_load
```
acroz: AFAICT this schema is only used to `dump` - this `post_load` method is never used so should be removed.

On a similar note, `required` is only respected on `load` and so we should probably be removed from the fields above as it's misleading.

tests/clients/test_environment.py
```diff
@@ -46,9 +139,95 @@
...
+    data = PythonSchema().load(PYTHON_BODY)
+    assert data == PYTHON
+
+
+def test_script_schema():
+    data = ScriptSchema().load(SCRIPT_BODY)
+    assert data == SCRIPT
+
+
+def test_specification_schema():
```
acroz: The `SpecificationSchema` (and all the schemas it depends on) is used on both `dump` and `load` - we should add tests for the `dump` case as well.

I suggest that for cases where we test both directions, we should rename the existing tests like:

```suggestion
def test_specification_schema_load():
```

and add additional tests named like `test_specification_schema_dump`.

faculty/clients/environment.py
```diff
@@ -30,9 +57,126 @@
...
+        "specification",
     ],
 )
 
+EnvironmentUpdate = namedtuple(
+    "EnvironmentUpdate", ["name", "description", "specification"]
+)
+
+
+class SemverSchema(BaseSchema):
```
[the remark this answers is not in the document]
ameroueh: This name choice comes from the fact that the GUI does some kind of validation (as far as I can tell, that the version contains three dots - I don't think it fully validates against semver). I was wondering whether the client should do the same (either no validation or full semver validation, in my opinion).

faculty/clients/environment.py
```diff
@@ -30,9 +57,126 @@
...
+    pip = fields.Nested(PipSchema(), required=True)
+
+    @post_load
+    def make_python_specification(self, data):
+        return PythonSpecification(**data)
+
+
+class PythonSchema(BaseSchema):
+    python_2 = fields.Nested(
+        PythonSpecificationSchema(), required=True, data_key="Python2"
```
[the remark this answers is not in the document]
ameroueh: The GUI seemed to implicitly require them - so I thought recreating the GUI expectation was the "safest" thing to do.

faculty/clients/environment.py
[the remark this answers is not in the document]
acroz: Sure, though we're doing no such validation here. The regex used by the backend, if you want to use it, is:

```python
    r"^(?:\d+\!)?\d+(?:\.\d+)*(?:(?:a|b|rc)\d+)?(?:\.post\d+)?(?:\.dev\d+)?$"
```

In any case we should make the model and schema have consistent names. My preference for naming is, in order:

- Version / VersionSchema
- SemanticVersion / SemanticVersionSchema
- Semver / SemverSchema

I don't know if Python package versions are actually enforced to be semantic versions, and I don't think it's especially relevant to users of the API, so again I have a strong preference for Version / VersionSchema.

faculty/clients/environment.py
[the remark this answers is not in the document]
acroz: The GUI always sends them because it was easier for the frontend team to implement it that way.

The backend it not required to send any of `Python2` or `Python3` in the response payload so it will be 'safer' to make them both optional.

faculty/clients/environment.py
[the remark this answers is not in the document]
acroz: On the regex, it's just occurred to me that you should check it actually gets used on dump before trying to include it. If it's only used on load I don't think it's valuable to validate the format of the string we get back from the API.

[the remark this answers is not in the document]
srstevenson: > I don't know if Python package versions are actually enforced to be semantic versions

It's not enforced, and several popular packages I know of (Twisted, pytz, and certifi) have adopted [CalVer](https://calver.org/) instead.

faculty/clients/environment.py
```diff
@@ -30,9 +57,126 @@
...
+class PythonSpecificationSchema(BaseSchema):
+    conda = fields.Nested(CondaSchema(), required=True)
+    pip = fields.Nested(PipSchema(), required=True)
+
+    @post_load
+    def make_python_specification(self, data):
+        return PythonSpecification(**data)
+
+
+class PythonSchema(BaseSchema):
```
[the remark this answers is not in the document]
ameroueh: There already is a `PythonSpecificationSchema` (used to deserialise the `Python2` and `Python3` keys). Not entirely happy with the current naming of `PythonSchema`, but not sure what a good alternative should be.

faculty/clients/environment.py
ameroueh: In the future I intend to use the client to load environment updates from, for example, JSON files. So I believe there is a use case for loading environment updates.

tests/clients/test_environment.py
```diff
@@ -46,9 +139,95 @@
...
+    data = SpecificationSchema().load(SPECIFICATION_BODY)
+    assert data == SPECIFICATION
+
+
+def test_environment_update_schema():
```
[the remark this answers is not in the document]
ameroueh: Eventually I woud like to let users load updates to their environments from JSON files. So I beliveve there is a use case for implementing `load()`.

faculty/clients/environment.py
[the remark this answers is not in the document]
acroz: In reality, the Python2 and Python3 keys are not exclusive - Python2 and Python3 are just the names of the default environment names - you can also specify others.

This may imply changing the current PythonSpecficationSchema to PythonEnvironmentSchema and/or removing the current PythonSchema and replacing it in SpecificationSchema to a `fields.Dict(fields.String(), fields.Nested(PythonSpecificationSchema()))`? Happy to discuss this in person to save time.

ameroueh:
Have implemented almost all suggestions - the notable exception are how to handle the possibilty that other environments than `Python2` or `Python3` are specified.

srstevenson:
> Have implemented almost all suggestions - the notable exception are is that I do not (yet) handle the possibilty that other environments than Python2 or Python3 are specified.

I've never come across anyone use that feature, so perhaps we choose to add that in a separate PR.

faculty/clients/environment.py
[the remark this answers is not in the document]
acroz: Ok, let's leave the structure as is, but my final suggestion would be to do the following renames:

- `PythonSpecificationSchema` -> `PythonEnvironmentSchema`
- `PythonSchema` -> `PythonSpecificationSchema`

As this communicates the intended meaning of the former to represent a Python conda environment.

tests/clients/test_environment.py
```diff
@@ -20,22 +20,115 @@
 from marshmallow import ValidationError
 
 from faculty.clients.environment import (
```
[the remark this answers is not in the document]
acroz: Bump @pbugnion @zblz @srstevenson.

tests/clients/test_environment.py
```diff
@@ -20,33 +20,302 @@
...
+    VersionSchema,
 )
 
+VERSION_BODY = {"constraint": "==", "identifier": "1.0.0"}
+VERSION = Version(constraint=Constraint.EQUAL, identifier="1.0.0")
+
+VERSION_BODY_LATEST = "latest"
+VERSION_LATEST = "latest"
+
+INVALID_VERSION_BODY = {"constraint": "==", "identifier": "not_semantic"}
```
acroz:
```suggestion
INVALID_VERSION_BODY = {"constraint": "==", "identifier": "invalid-identifier"}
```
as being SemVer or not isn't relevant here.

tests/clients/test_environment.py
```diff
@@ -20,33 +20,302 @@
...
+INVALID_VERSION = Version(
+    constraint=Constraint.EQUAL, identifier="not_semantic"
```
acroz:
```suggestion
    constraint=Constraint.EQUAL, identifier="invalid-identifier"
```
as above.

tests/clients/test_environment.py
```diff
@@ -20,33 +20,302 @@
...
+    data = AptSchema().load(APT_BODY)
+    assert data == APT
+
+
+def test_apt_schema_dump():
+    data = AptSchema().dump(APT)
+    assert data == APT_BODY
+
+
+def test_python_schema_load():
```
acroz: Can we also test load and dump in the case that Python2 and/or Python3 are unset (None / not present in the BODY dict)? I'm not sure that would work at the moment (I think the fields in the `PythonSchema` may require a `missing=None` argument each).

tests/clients/test_environment.py
```diff
@@ -70,3 +339,90 @@ def test_environment_client_list(mocker):
...
+
+    schema_mock.assert_called_once_with()
+    EnvironmentClient._get.assert_called_once_with(
+        "/project/{}/environment/{}".format(PROJECT_ID, ENVIRONMENT_ID),
+        schema_mock.return_value,
+    )
+
+
+def test_environment_client_update(mocker):
+    mocker.patch.object(EnvironmentClient, "_put_raw", return_value=None)
```
acroz:
```suggestion
    mocker.patch.object(EnvironmentClient, "_put_raw")
```
Don't need this as the return value is not used.

tests/clients/test_environment.py
```diff
@@ -70,3 +339,90 @@ def test_environment_client_list(mocker):
...
+    environment_update_create_mock = mocker.patch(
```
acroz: I don't think it's worthwhile to mock out the instantiation of a `namedtuple`. Can we remove this mock and just check that a `EnvironmentCreateUpdate` object corresponding to the inputs passed to the function is passed to `EnvironmentCreateUpdateSchema.dump`?

tests/clients/test_environment.py
```diff
@@ -70,3 +339,90 @@ def test_environment_client_list(mocker):
...
+        "/project/{}/environment/{}".format(PROJECT_ID, ENVIRONMENT_ID),
+        json=ENVIRONMENT_CREATE_UPDATE_BODY,
+    )
+
+
+def test_environment_client_create(mocker):
+    mocker.patch.object(
+        EnvironmentClient, "_post", return_value=ENVIRONMENT_CREATION_RESPONSE
+    )
+    environment_update_create_mock = mocker.patch(
```
acroz: As above, I'd recommend not mocking this.

tests/clients/test_environment.py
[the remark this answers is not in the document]
srstevenson: I have an aversion to wildcard imports, but will side with whatever the consensus is in this case.

pbugnion:
I think this looks great! Thanks Ares -- really useful.

acroz:
This is released with 0.23.4.

PR #102, opened 2019-04-29
Change delete_runs/restore_runs in experiment client to use the new filtered endpoint

tomas-milata:
Also, delete/restore no runs if [] is passed. It seems intuitive to not delete/restore anything if `[]` is passed and this makes it consistent with `list_runs`.

This can be merged tomorrow (when the new backend is in place).

Changing the behaviour for the empty list is itself a breaking change, but I'd suggest not bumping minor - in my mind it's a bugfix - hopefully no one was relying on this behaviour, or do people have other thoughts?

### DIFF
--- a/faculty/clients/experiment.py
+++ b/faculty/clients/experiment.py
@@ -628,12 +628,14 @@ class ExperimentClient(BaseClient):
         """
         endpoint = "/project/{}/run/delete/query".format(project_id)
 
-        payload = {}
-        if run_ids is not None:
-            if len(run_ids) == 0:
-                return DeleteExperimentRunsResponse(
-                    deleted_run_ids=[], conflicted_run_ids=[]
-                )
+        if run_ids is None:
+            # Delete all runs in project
+            payload = {}  # No filter
+        elif len(run_ids) == 0:
+            return DeleteExperimentRunsResponse(
+                deleted_run_ids=[], conflicted_run_ids=[]
+            )
+        else:
             payload = {
                 "filter": {
                     "operator": "or",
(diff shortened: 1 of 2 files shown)


eliasbenussi:
Personally, I think the change you made for `run_ids=[]` is ok. I think that was the de facto behaviour, and here you are just short circuiting?

I haven't approved to give others time to review, but in principle this LGTM

faculty/clients/experiment.py
```diff
@@ -645,24 +656,33 @@ def restore_runs(self, project_id, run_ids=None):
...
         Returns
         -------
         RestoreExperimentRunsResponse
             Containing lists of successfully restored and conflicting (already
             active) run IDs.
         """
+        endpoint = "/project/{}/run/restore/query".format(project_id)
 
-        query_params = []
+        payload = {}
```
acroz: What would you think about having something like:

```python
if run_ids is None:
    # Delete all runs in project
    payload = {}  # No filter
elif len(run_ids) == 0:
    return RestoreExperimentRunsResponse(
        restored_run_ids=[], conflicted_run_ids=[]
    )
else:
    payload = {
        "filter": {
            "operator": "or",
            "conditions": [
                {"by": "runId", "operator": "eq", "value": str(run_id)}
                for run_id in run_ids
            ],
        }
    }
```

This would remove a level of nesting plus make it more explicit how `run_ids=None` results in deleting all runs of the project.

tomas-milata:
> Personally, I think the change you made for `run_ids=[]` is ok. I think that was the de facto behaviour, and here you are just short circuiting?
> 

The previous implementation would *delete all* runs for `run_ids=[]` - it would result in no query parameters -> "no filter" -> "delete all in project". So it's not just short circuiting, it's also changing the behaviour.

faculty/clients/experiment.py
tomas-milata: 👍

PR #103, opened 2019-04-29
Return full serialised server on server creation

srstevenson:
This is to allow the CLI to print the server name in addition to the ID, as proposed in https://github.com/facultyai/faculty-cli/pull/15.

### DIFF
--- a/faculty/clients/server.py
+++ b/faculty/clients/server.py
@@ -189,9 +189,10 @@ class ServerClient(BaseClient):
                 str(env_id) for env_id in initial_environment_ids
             ]
 
-        return self._post(
+        server_id = self._post(
             "/instance/{}".format(project_id), ServerIdSchema(), json=payload
         )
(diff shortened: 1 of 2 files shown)


srstevenson:
I agree with the convention: I'd prefer that the Galleon endpoint return the serialised server resource itself so only a single request is required, whether that be here or in the CLI, however pushing through that change to Galleon now seems unlikely.

Taken that as given, I'm happy to add the additional call to `.get()` in the CLI instead, as it's only the ability to print the server name after creation I'm interested in so I can type `faculty shell <project> <server name>` immediately afterwards; `faculty shell <project> <server id>` is a lot to type, or requires copying and pasting.

acroz:
That sounds like a good way forward - I think that will be a good improvement to the CLI.

Closing this PR as we've agreed not to make the changes contained here.

acroz:
@srstevenson do you mind if I delete the branch here as well?

PR #107, opened 2019-05-13
Add query runs method

eliasbenussi:


faculty/clients/experiment.py
```diff
@@ -99,6 +106,72 @@ class ExperimentRunStatus(Enum):
...
+class CompoundFilterOperator(Enum):
+    AND = "and"
+    OR = "or"
+
+
+_Sort = namedtuple("_Sort", ["by", "key", "order"])
+
+class Sort(_Sort):
+    def __new__(self, by, key, order):
+        if by in [SortBy.STARTED_AT, SortBy.RUN_NUMBER, SortBy.DURATION] and key is not None:
```
srstevenson:
```suggestion
        if by in {SortBy.STARTED_AT, SortBy.RUN_NUMBER, SortBy.DURATION} and key is not None:
```

faculty/clients/experiment.py
```diff
@@ -99,6 +106,72 @@ class ExperimentRunStatus(Enum):
...
+            raise ValueError("key must be none for type {}".format(by))
+        elif by in [SortBy.TAG, SortBy.PARAM, SortBy.METRIC] and key is None:
```
srstevenson:
```suggestion
        elif by in {SortBy.TAG, SortBy.PARAM, SortBy.METRIC} and key is None:
```

faculty/clients/experiment.py
```diff
@@ -245,6 +330,93 @@ def make_restore_runs_response(self, data):
...
+            or obj.by
+            in {
+                SingleFilterBy.TAG,
+                SingleFilterBy.PARAM,
+                SingleFilterBy.METRIC,
+            }
+        )
+
+    def _deserialize(self, value, attr, obj, **kwargs):
+        pass
```
eliasbenussi: Not really sure what the sensible thing to do is here. This is strictly a request payload, which I don't predict we will need to deserialise any time soon.

tests/clients/test_experiment.py
```diff
@@ -329,6 +340,241 @@ def test_experiment_run_data_schema_multiple():
...
+        SingleFilter(
+            SingleFilterBy.TAG,
+            "tag_key",
+            SingleFilterOperator.EQUAL_TO,
+            "tag_value",
+        ),
+        None,
+    ],
+)
+OR_FILTER_BODY = {
```
tomas-milata: This seems unused.

tests/clients/test_experiment.py
```diff
@@ -329,6 +340,241 @@ def test_experiment_run_data_schema_multiple():
...
+            "by": "tag",
+            "key": "tag_key",
+            "operator": "eq",
+            "value": "tag_value",
+        },
+        None,
+    ],
+}
+
+OR_FILTER = CompoundFilter(
```
tomas-milata: If I understand correctly, the structure of filters is recursive, allows arbitrary level of nesting (at least the backend supports it). Is it worth having a test case with at least 2 levels? E.g. something like `OR(F1, AND(F2, F3))`.

tomas-milata:
It'd be nice to change `delete_run` and `restore_runs` to use the new schema too. I appreciate it's extra work, but I think it's worth it for the sake of consistency.

tomas-milata:
Did you consider using a library for the polymorphic filter type? E.g. https://github.com/marshmallow-code/marshmallow-oneofschema or https://github.com/Bachmann1234/marshmallow-polyfield. (I'm not saying we need to use it).

faculty/clients/experiment.py
```diff
@@ -532,6 +708,50 @@ def list_runs(
...
+        until ``next`` is returned as ``None``.
+
+        Parameters
+        ----------
+        project_id : uuid.UUID
+        filter: either SingleFilter object or CompoundFilter object, optional
+            To filter runs of experiments with the given filter. By default,
+            runs from all experiments are returned.
+        sort: List[Sort], optional
+            Runs are ordered using sorting elements lexicographically. By
```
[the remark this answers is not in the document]
eliasbenussi: My understanding was that if a user passes a list of Sort, then the relative importance of each condition was gradually decreasing? (first Sort object would take precedence over the second, if first is the same than consider second element, etc)?

eliasbenussi:
> It'd be nice to change `delete_run` and `restore_runs` to use the new schema too. I appreciate it's extra work, but I think it's worth it for the sake of consistency.

@tomas-milata I agree that we should do it, but I was considering doing it in a separate PR. My reasoning is that the `query_runs` function we implemented was required in @acroz 's PR (https://github.com/facultyai/faculty/pull/106) and so I would just like to get it out as soon as it's ready.

In fact I would say that, other than `delete_runs` and `restore_runs`, we could also reimplement `list_runs` to internally call `query_runs` as it is just a special case at this point -- although possibly a convenient one to have

eliasbenussi:
> Did you consider using a library for the polymorphic filter type? E.g. https://github.com/marshmallow-code/marshmallow-oneofschema or https://github.com/Bachmann1234/marshmallow-polyfield. (I'm not saying we need to use it).

@tomas-milata I did not, but I after taking a look I think that:

* the first one leads to a decoding that is unfortunately not correct for our payload (it explicitly specifies the type of the object). It adds a `type` field (although I think this can be removed), and it names the key for the value according to its type. Neither of these things are what we want.

* the second one still requires to explicitly implement serialisation and deserialisation functions. This is verbose enough that I think our current implementation is basically just as good, with the advantage of one less dependency

eliasbenussi:
> > Did you consider using a library for the polymorphic filter type? E.g. https://github.com/marshmallow-code/marshmallow-oneofschema or https://github.com/Bachmann1234/marshmallow-polyfield. (I'm not saying we need to use it).
> 
> @tomas-milata I did not, but I after taking a look I think that:
> 
> * the first one leads to a decoding that is unfortunately not correct for our payload (it explicitly specifies the type of the object). It adds a `type` field (although I think this can be removed), and it names the key for the value according to its type. Neither of these things are what we want.
> * the second one still requires to explicitly implement serialisation and deserialisation functions. This is verbose enough that I think our current implementation is basically just as good, with the advantage of one less dependency

Actually I think my points on the first one are wrong. That one might be a viable solution

tests/clients/test_experiment.py
```diff
@@ -329,6 +341,277 @@ def test_experiment_run_data_schema_multiple():
...
+METRIC_FILTER_BODY = {
+    "by": "metric",
+    "key": "metric_key",
+    "operator": "eq",
+    "value": "metric_value",
+}
+
+AND_FILTER = CompoundFilter(
+    operator=CompoundFilterOperator.AND,
+    conditions=[
```
acroz: To shorten this (and `AND_FILTER_BODY`, `OR_FILTER`, etc.) you could just use the already existing single filters like `TAG_FILTER` to compose compound filters.

faculty/clients/experiment.py
```diff
@@ -245,6 +334,98 @@ def make_restore_runs_response(self, data):
...
+    def _deserialize(self, value, attr, data, **kwargs):
+        return value
+
+    def _serialize(self, value, attr, obj, **kwargs):
+        if self._is_directly_stringifiable(value, obj):
+            return str(value)
+        elif obj.by == SingleFilterBy.DELETED_AT:
+            return marshmallow_utils.from_iso_datetime(str(value)).isoformat()
+        else:
+            raise RunQueryFilterValidation(
```
acroz: Why not just raise a marshmallow `ValidationError`? I feel this would be more idiomatic.

faculty/clients/experiment.py
```diff
@@ -245,6 +334,98 @@ def make_restore_runs_response(self, data):
...
+            return SingleFilterSchema().load(value)
+        else:
+            return CompoundFilterSchema().load(value)
+
+    def _serialize(self, value, attr, obj, **kwargs):
+        print(type(obj))
+        if value is None and isinstance(obj, QueryRuns):
+            return None
+        elif value is None:
+            raise RunQueryFilterValidation(
```
acroz: As above:

> Why not just raise a marshmallow `ValidationError`? I feel this would be more idiomatic.

faculty/clients/experiment.py
```diff
@@ -532,6 +713,50 @@ def list_runs(
...
+        Page(start=10, limit=10)
+
+        Get all experiment runs by making successive calls to ``query_runs``,
+        passing the ``start`` and ``limit`` of the ``next`` page each time
+        until ``next`` is returned as ``None``.
+
+        Parameters
+        ----------
+        project_id : uuid.UUID
+        filter: either SingleFilter object or CompoundFilter object, optional
```
acroz:
```suggestion
        filter: SingleFilter or CompoundFilter, optional
```

faculty/clients/experiment.py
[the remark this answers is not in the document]
acroz: Can we just replace 'lexographically' with some simpler language?

faculty/clients/experiment.py
```diff
@@ -245,6 +334,98 @@ def make_restore_runs_response(self, data):
         return RestoreExperimentRunsResponse(**data)
 
 
+class SingleFilterValueField(fields.Field):
+    """
+    Field that serialises/deserialises a run filter.
+    """
+
+    def _is_valid_uuid(self, value, obj):
+        return isinstance(value, uuid.UUID) and (
```
acroz: marshmallow allows you to pass strings for UUIDs, provided they are valid UUIDs. It would be good to preserver this behaviour.

Could we be cheeky and do something like:

```python
class SingleFilterValueField(fields.Field):

    def _serialize(self, value, attr, obj, **kwargs):
        if obj.by in {SingleFilterBy.PROJECT_ID, SingleFilterBy.RUN_ID}:
            field = fields.UUID()
        elif obj.by = SingleFilterBy.EXPERIMENT_ID:
            field = fields.Integer()
        elif obj.by = SingleFilterBy.DELETED_AT:
            field = fields.DateTime()
        # etc. for tag, param, metric
        return field._serialize(value, attr, obj, **kwargs)
```

faculty/clients/experiment.py
```diff
@@ -245,6 +334,98 @@ def make_restore_runs_response(self, data):
...
+
+    def _is_directly_stringifiable(self, value, obj):
+        return (
+            self._is_valid_uuid(value, obj)
+            or self._is_valid_experiment_id(value, obj)
+            or obj.by
+            in {
+                SingleFilterBy.TAG,
+                SingleFilterBy.PARAM,
+                SingleFilterBy.METRIC,
```
acroz: Metrics need to be passed as numbers and params can be either strings or numbers.

faculty/clients/experiment.py
```diff
@@ -245,6 +334,98 @@ def make_restore_runs_response(self, data):
...
+    def _deserialize(self, value, attr, data, **kwargs):
+        if value is None:
+            return None
+        elif isinstance(value, SingleFilter):
+            return SingleFilterSchema().load(value)
+        else:
+            return CompoundFilterSchema().load(value)
+
+    def _serialize(self, value, attr, obj, **kwargs):
+        print(type(obj))
```
acroz: Remove this line?

faculty/clients/experiment.py
```diff
@@ -245,6 +334,98 @@ def make_restore_runs_response(self, data):
...
+        if value is None and isinstance(obj, QueryRuns):
```
acroz: Why do we need to check the type of the obj here?

faculty/clients/experiment.py
zblz: In general, marshmallow does not validate on serialization, only on deserialization, so it's not obvious tha a `ValidationError` is the correct thing here as it is (by use, if not by definiton) only applicable ot deserialization.

eliasbenussi: To be honest I don't really know what the right thing to do is here. Elsewhere in my additions I have just used a `ValueError` (e.g. in the constructors of Filters). Here I wanted to highlight that it's a serialisation error, but I am happy to change it if anyone has strong thoughts

faculty/clients/experiment.py
eliasbenussi: I had thought about doing something like this, but I worried about breaking conceptual encapsulation? Even though, to be fair, I don't understand why they would consider an explicit call to `_serialize` (or `_deserialize`) a bad thing.

If you believe it's okay to call `_serialize` I am happy to go for it

faculty/clients/experiment.py
eliasbenussi: It's because of the recursive nature of this serialisation:
- QueryRuns is allowed to have None as a filter
- However, compound filters are not allowed to have None filters in their conditions

faculty/clients/experiment.py
acroz: I think it's fine to use it here, especially since the API to `_serialize` is stable / supported (we're implementing it ourselves for `SingleFilterValueField`.

faculty/clients/experiment.py
```diff
@@ -245,6 +350,113 @@ def make_restore_runs_response(self, data):
...
+class SingleFilterSchema(BaseSchema):
+    by = EnumField(SingleFilterBy, by_value=True, required=True)
+    key = fields.String()
+    operator = EnumField(SingleFilterOperator, by_value=True, required=True)
+    value = SingleFilterValueField(required=True)
+
+
+class CompoundFilterSchema(BaseSchema):
+    operator = EnumField(CompoundFilterOperator, by_value=True, required=True)
+    conditions = fields.List(FilterField(skip_if=None))
```
acroz: What does `skip_if` do? I can't find any mention of it in the documentation.

haileyfong: I think this was added by mistake. Thanks for spotting this out, I will remove it now.

PR #109, opened 2019-05-29
Add step to metric 

haileyfong:
Allow users to specify step in metric. If users did not set it, step is None by default.


faculty/clients/experiment.py
```diff
@@ -83,7 +83,8 @@ class ExperimentRunStatus(Enum):
     ],
 )
 
-Metric = namedtuple("Metric", ["key", "value", "timestamp"])
+Metric = namedtuple("Metric", ["key", "value", "timestamp", "step"])
+Metric.__new__.__defaults__ = (None,)
```
zblz: It took me a while to realize that this sets the default value for the *last* field of the namedtuple, as opposed to the first one. I'm still not sure why it does that, and the pattern of using `__new__.__defaults__` is not mentioned in the docs, so I would refrain from using it.

A better (IMO) pattern would be the one given in [this stackoverflow answer](https://stackoverflow.com/a/16721002) because it exposes that `step` is the optional attribute with a default value of `None` in the `Metric` class:

```python
class Metric(namedtuple("Metric", ["key", "value", "timestamp", "step"])):
    __slots__ = ()
    def __new__(cls, key, value, timestamp, step=None):
        return super(Node, cls).__new__(cls, key, value, timestamp, step)
```

zblz: Also, it's not obvious to me that having a default is necessary at all. The marshmallow schema below will already fill in `None` if the response from the service does not have a step, so the actual namedtuple can require having `step` set.

tests/clients/test_experiment.py
```diff
@@ -105,33 +106,58 @@
...
+}
+
+METRIC_WITHOUT_STEP = Metric(
+    key=METRIC_KEY, value=123, timestamp=METRIC_TIMESTAMP
+)
+METRIC_WITHOUT_STEP_BODY = {
+    "key": METRIC_KEY,
+    "value": 123.0,
+    "timestamp": METRIC_TIMESTAMP_STRING_PYTHON,
+    "step": None,
```
zblz: This key should not be here at all to test marshmallow adding the `None` default if missing.

tests/clients/test_experiment.py
```diff
@@ -105,33 +106,58 @@
...
+METRIC = Metric(key=METRIC_KEY, value=123, timestamp=METRIC_TIMESTAMP, step=0)
 METRIC_BODY = {
     "key": METRIC_KEY,
     "value": 123.0,
     "timestamp": METRIC_TIMESTAMP_STRING_PYTHON,
+    "step": 0,
+}
+
+METRIC_WITHOUT_STEP = Metric(
+    key=METRIC_KEY, value=123, timestamp=METRIC_TIMESTAMP
```
zblz: If we make step required as per above comment, this should have `step=None`:
```suggestion
    key=METRIC_KEY, value=123, timestamp=METRIC_TIMESTAMP, step=None
```

tests/clients/test_experiment.py
```diff
@@ -662,7 +699,8 @@ def test_experiment_client_list_runs_page(mocker):
     )
 
 
-def test_log_run_data(mocker):
+@pytest.mark.parametrize("metric", [METRIC, METRIC_WITHOUT_STEP])
```
zblz: Is this necessary? We are just testing that `ExperimentRunDataSchema` has been called, not that it has done anything with the input, so `METRIC` and `METRIC_WITHOUT_STEP` are functionally equal in this test.

zblz:
> Allow users to specify step in metric. If users did not set it, step is zero by default.

Is this right? According to the code, the step is set to `None`, not zero.

faculty/clients/experiment.py
eliasbenussi: We have that default in `Metric` because of its use in mlflow-faculty. At the moment mlflow (0.9) and mlflow-faculty don't have the concept of step. This default allows it to work with auto-increasing steps in the backend set by atlas.

When we upgrade to mlflow 1.0, we should also update mlflow-faculty. At that point in mlflow they send step=0  by default (incidentally the reason for that error in the docs string)

tests/clients/test_experiment.py
eliasbenussi: Ditto as per response above

haileyfong:
> > Allow users to specify step in metric. If users did not set it, step is zero by default.
> 
> Is this right? According to the code, the step is set to `None`, not zero.

You are right, the description was not updated when the default value of step was changed to None.

tests/clients/test_experiment.py
```diff
@@ -105,33 +106,52 @@
...
-METRIC = Metric(key=METRIC_KEY, value=123, timestamp=METRIC_TIMESTAMP)
+METRIC = Metric(key=METRIC_KEY, value=123, timestamp=METRIC_TIMESTAMP, step=0)
 METRIC_BODY = {
     "key": METRIC_KEY,
     "value": 123.0,
     "timestamp": METRIC_TIMESTAMP_STRING_PYTHON,
+    "step": 0,
+}
+
+METRIC_WITHOUT_STEP = Metric(
```
acroz: Could remove this case now that we're no longer expecting a situation where 'missing' steps are sent.

tests/clients/test_experiment.py
```diff
@@ -105,33 +106,52 @@
...
 
 METRIC_SECOND_TIMESTAMP = datetime(2018, 3, 12, 16, 20, 30, 122000, tzinfo=UTC)
 METRIC_SECOND_TIMESTAMP_STRING_PYTHON = "2018-03-12T16:20:30.122000+00:00"
 METRIC_SECOND_MEASURE = Metric(
-    key=METRIC_KEY, value=127, timestamp=METRIC_SECOND_TIMESTAMP
+    key=METRIC_KEY, value=127, timestamp=METRIC_SECOND_TIMESTAMP, step=0
+)
+
+METRIC_HISTORY = MetricHistory(
+    3, False, METRIC_KEY, [METRIC, METRIC_SECOND_MEASURE]
```
acroz: To make consistent with the contained data:
```suggestion
    2, False, METRIC_KEY, [METRIC, METRIC_SECOND_MEASURE]
```

tests/clients/test_experiment.py
```diff
@@ -105,33 +106,52 @@
...
 )
 
-METRIC_HISTORY = [METRIC, METRIC_SECOND_MEASURE]
 METRIC_HISTORY_BODY = {
+    "totalEntriesNumber": 3,
```
acroz: As above:
```suggestion
    "totalEntriesNumber": 2,
```

tests/clients/test_experiment.py
```diff
@@ -105,33 +106,52 @@
...
+    "value": 123.0,
+    "timestamp": METRIC_TIMESTAMP_STRING_PYTHON,
+    "step": None,
 }
 
 METRIC_SECOND_TIMESTAMP = datetime(2018, 3, 12, 16, 20, 30, 122000, tzinfo=UTC)
 METRIC_SECOND_TIMESTAMP_STRING_PYTHON = "2018-03-12T16:20:30.122000+00:00"
 METRIC_SECOND_MEASURE = Metric(
-    key=METRIC_KEY, value=127, timestamp=METRIC_SECOND_TIMESTAMP
+    key=METRIC_KEY, value=127, timestamp=METRIC_SECOND_TIMESTAMP, step=0
```
acroz: To make this case more different from `METRIC`:
```suggestion
    key=METRIC_KEY, value=127, timestamp=METRIC_SECOND_TIMESTAMP, step=1
```

tests/clients/test_experiment.py
```diff
@@ -105,33 +106,52 @@
...
             "key": METRIC_KEY,
             "value": 123.0,
             "timestamp": METRIC_TIMESTAMP_STRING_PYTHON,
+            "step": 0,
         },
         {
             "key": METRIC_KEY,
             "value": 127.0,
             "timestamp": METRIC_SECOND_TIMESTAMP_STRING_PYTHON,
+            "step": 0,
```
acroz: As above:
```suggestion
            "step": 1,
```

tests/clients/test_experiment.py
```diff
@@ -297,6 +317,11 @@ def test_metric_schema():
     assert data == METRIC
 
 
+def test_validation_metric_schema():
```
acroz: There's no longer a reason to treat step differently on load. I suggest removing this test.

tests/clients/test_experiment.py
```diff
@@ -312,11 +337,18 @@ def test_tag_schema_dump():
     assert data == TAG_BODY
 
 
-def test_experiment_run_data_schema():
+@pytest.mark.parametrize(
+    "metric,metric_body",
```
acroz: Nit:
```suggestion
    "metric, metric_body",
```

tests/clients/test_experiment.py
```diff
@@ -312,11 +337,18 @@ def test_tag_schema_dump():
...
+    [[METRIC, METRIC_BODY], [METRIC_WITHOUT_STEP, METRIC_WITHOUT_STEP_BODY]],
```
acroz: I find having the argument sets as tuples aids readability and better reflects their having a well defined length:
```suggestion
    [(METRIC, METRIC_BODY), (METRIC_WITHOUT_STEP, METRIC_WITHOUT_STEP_BODY)],
```

tests/clients/test_experiment.py
```diff
@@ -312,11 +337,18 @@ def test_tag_schema_dump():
...
+)
+def test_experiment_run_data_schema(metric, metric_body):
     data = ExperimentRunDataSchema().dump(
-        {"metrics": [METRIC], "params": [PARAM], "tags": [TAG]}
+        {"metrics": [metric], "params": [PARAM], "tags": [TAG]}
     )
-    assert data == EXPERIMENT_RUN_DATA_BODY
+
+    EXPECTED_EXPERIMENT_RUN_DATA_BODY = EXPERIMENT_RUN_DATA_BODY
+    EXPECTED_EXPERIMENT_RUN_DATA_BODY["metrics"] = [metric_body]
```
acroz: This code actually modifies the module-level variable. The line above is just making a second reference to the same (mutable) dictionary, which we are then mutating with the second line.

I suggest something like:

```python
    expected = EXPERIMENT_RUN_DATA_BODY.copy()
    expected["metrics"] = [metric_body]
```

Note that I've made the variable non-caps as that's a Python convention meant to indicate that a variable should be treated as a constant (Python does not have actual constants encoded in the language).

acroz: For the sake of clarity, I think that my suggestion above would not work if mutating some more deeply nested data structure. From what I remember, `.copy()` only makes a 'shallow copy' of the dictionary, which does not actually copy the contents of any collections inside (it makes a reference).

eliasbenussi: Yeah sorry you are right. As we were writing this code I told myself to double check this was fine, because I know it wasn't very idiomatic, and then forgot. Good catch

PR #110, opened 2019-06-11
Refactor filter and sort objects and associated schemas

acroz:
I didn't like how we were doing validation of Filter contents in a bunch of different places, so did some refactoring including the following changes:

* Each filter and sort type is now encoded with its own type, making it clearer when it takes a key or not
* Each filter and sort type has an associated schema, making it much clearer what value types are permitted
* All validation logic is now at serialisation time, not mixed between serialisation time and construction of the objects - this feels more idiomatic for Python + marshmallow

While updating test, a number have been improved, and the implementation of 2 (presumably untested) methods has been fixed.

In a subsequent PR, I'd like to split the experiments module and tests into multiple files to aid readability.

### DIFF
--- a/tests/clients/test_experiment.py
+++ b/tests/clients/test_experiment.py
@@ -22,49 +22,58 @@ from pytz import UTC
 
 from faculty.clients.base import Conflict
 from faculty.clients.experiment import (
-    CreateRunSchema,
+    ComparisonOperator,
     CompoundFilter,
-    CompoundFilterOperator,
+    CreateRunSchema,
     DeleteExperimentRunsResponse,
     DeleteExperimentRunsResponseSchema,
+    DeletedAtFilter,
+    DurationSort,
     Experiment,
     ExperimentClient,
-    ExperimentNameConflict,
     ExperimentDeleted,
+    ExperimentIdFilter,
+    ExperimentNameConflict,
     ExperimentRun,
     ExperimentRunDataSchema,
     ExperimentRunSchema,
     ExperimentRunStatus,
     ExperimentSchema,
+    FilterSchema,
     LifecycleStage,
     ListExperimentRunsResponse,
     ListExperimentRunsResponseSchema,
+    LogicalOperator,
     Metric,
     MetricDataPoint,
-    MetricSchema,
+    MetricFilter,
     MetricHistory,
     MetricHistorySchema,
+    MetricSchema,
+    MetricSort,
     Page,
     PageSchema,
     Pagination,
     PaginationSchema,
     Param,
     ParamConflict,
+    ParamFilter,
     ParamSchema,
-    QueryRuns,
-    QueryRunsSchema,
+    ParamSort,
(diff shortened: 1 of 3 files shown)


tests/clients/test_experiment.py
```diff
@@ -348,334 +357,315 @@ def test_experiment_run_data_schema_multiple():
...
+
+
+@pytest.mark.parametrize(
+    "filter_type, value", [(ParamFilter, None), (MetricFilter, "invalid")]
+)
+def test_filter_schema_invalid_value_with_key(filter_type, value):
+    filter = filter_type("key", ComparisonOperator.EQUAL_TO, value)
+    with pytest.raises(ValidationError):
+        FilterSchema().dump(filter)
+
```
haileyfong: Should we also test the situation when key is absent for tag, param and metric filters?

tests/clients/test_experiment.py
```diff
@@ -348,334 +357,315 @@ def test_experiment_run_data_schema_multiple():
...
     ],
 )
 @pytest.mark.parametrize(
-    "order, order_body", [(SortOrder.ASC, "asc"), (SortOrder.DESC, "desc")]
+    "order, expected_order", [(SortOrder.ASC, "asc"), (SortOrder.DESC, "desc")]
 )
-def test_sort_schema(by, key, by_body, order, order_body):
-    sort = Sort(by, key, order)
-    expected_json = {"by": by_body, "key": key, "order": order_body}
+def test_sort_schema_no_tag(sort_type, by, order, expected_order):
```
haileyfong: Is it better to name it as test_sort_schema_no_key instead of test_sort_schema_no_tag, because we have been using no key in the filter tests, also, the key field is named key on construction of a sort instance.

tests/clients/test_experiment.py
```diff
@@ -348,334 +357,315 @@ def test_experiment_run_data_schema_multiple():
...
-    with pytest.raises(ValueError, match="key must be none"):
-        Sort(SortBy.RUN_NUMBER, "invalid_number", SortOrder.ASC)
+@pytest.mark.parametrize(
+    "sort_type, by",
+    [(TagSort, "tag"), (ParamSort, "param"), (MetricSort, "metric")],
+)
+@pytest.mark.parametrize(
+    "order, expected_order", [(SortOrder.ASC, "asc"), (SortOrder.DESC, "desc")]
+)
+def test_sort_schema_with_tag(sort_type, by, order, expected_order):
```
haileyfong: same as above

tests/clients/test_experiment.py
acroz: Ah, this was a mistake, good spot.

tests/clients/test_experiment.py
acroz: I don't think this is necessary as the namedtuple constructor will require users to pass a key for these filter types.

PR #114, opened 2019-06-13
Add server context helper

acroz:
This provides a convenience interface for getting information about the server context that will be used by the resource-based interface in a subsequent PR.

### DIFF
--- a/faculty/context.py
+++ b/faculty/context.py
@@ -52,7 +52,7 @@ def _get_environ_as_type(key, cls):
     except ValueError:
         # Badly formatted
         template = (
-            "Error interpreting badly formatted environment varaible {}={}"
+            "Error interpreting badly formatted environment variable {}={}"
         )
         warnings.warn(template.format(key, os.environ[key]))
         return None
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -33,6 +33,7 @@
         "enum34; python_version<'3.4'",
         # Install marshmallow with 'reco' (recommended) extras to ensure a
         # compatible version of python-dateutil is available
+        "attr",
```
zblz: `attr` is, confusingly, another package:
```suggestion
        "attrs",
```

faculty/context.py
```diff
@@ -0,0 +1,83 @@
...
+def _get_environ_as_type(key, cls):
+    try:
+        return cls(os.environ[key])
+    except KeyError:
+        # Not in environment
+        return None
+    except ValueError:
+        # Badly formatted
+        template = (
+            "Error interpreting badly formatted environment varaible {}={}"
```
srstevenson:
```suggestion
            "Error interpreting badly formatted environment variable {}={}"
```

acroz: thanks 👍 - good spot

setup.py
acroz: Well spotted - this only worked in tests as `pytest` depends on `attrs`.

PR #116, opened 2019-06-17
Pin version of marshmallow

haileyfong:
The newest version of marshmallow introduced backwards incompatible changes that are incompatible with our code:
-  Include key-word arguments ```many``` and ```partial``` in decorated methods e.g. ```pre_load```, ```post_load```, etc. 

Therefore, many tests are not passing because of this. Pinning the version of marshmallow to an older version could solve this problem for the time being.


haileyfong:
All tests on travis-ci passed except the licence header check.

PR #120, opened 2019-07-05
Fix adding apt packages when creating environments

pbugnion:
Currently, any attempt to include apt packages in the environment specification fails. This is because the client specifies packages as:

```json
{ "name": "..." }
```

whereas Baskerville wants:

```json
{ "name": "...", "version": "VERSION_IDENTIFIER" }
```

where `VERSION_IDENTIFIER` is `"latest" | {"constraint": "==", "identifier": "1.2.3" }`.

For a minimal example:

```py
import faculty
import os

ENVIRONMENT_SPECIFICATION = {
    "apt": {
        "packages": [
            {"name": "tree", "version": "latest"},
        ]
    },
    "python": {
        "python2": {
            "conda": {"packages": [], "channels": []},
            "pip": {"packages": [], "extra_index_urls": []}
        },
        "python3": {
            "conda": {"packages": [], "channels": []},
            "pip": {"packages": [], "extra_index_urls": []}
        }
    },
    "bash": {
        "script": "echo hello"
    }
}

c = faculty.client("environment")
c.create(os.environ["FACULTY_PROJECT_ID"], "test-544", ENVIRONMENT_SPECIFICATION, description="hello")
```

### DIFF
--- a/faculty/clients/environment.py
+++ b/faculty/clients/environment.py
@@ -40,7 +40,7 @@ Pip = namedtuple("Pip", ["extra_index_urls", "packages"])
 Conda = namedtuple("Conda", ["channels", "packages"])
 PythonEnvironment = namedtuple("PythonEnvironment", ["pip", "conda"])
 Apt = namedtuple("Apt", ["packages"])
-AptPackage = namedtuple("AptPackage", ["name"])
+AptPackage = namedtuple("AptPackage", ["name", "version"])
 Script = namedtuple("Script", ["script"])
 PythonSpecification = namedtuple("PythonSpecification", ["python2", "python3"])
 Specification = namedtuple("Specification", ["apt", "bash", "python"])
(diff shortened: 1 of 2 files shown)


faculty/clients/environment.py
```diff
@@ -62,18 +63,49 @@ class Constraint(Enum):
...
+
+    @post_dump
+    def dump_version(self, data):
+        self.validate_version_format(data["identifier"])
+        return data
+
+
+class VersionSchema(PythonVersionSchema):
+    def __init__(self, *args, **kwargs):
+        template = (
```
acroz: Would happliy also just remove this, as I think it's extremely unlikely anyone will be depending on the `VersionSchema` directly.

If it stays, suggest:
```suggestion
        message = (
```

faculty/clients/environment.py
```diff
@@ -95,18 +127,46 @@ def _deserialize(self, value, attr, obj, **kwargs):
...
+    def _serialize(self, value, attr, obj, **kwargs):
+        if value == "latest":
+            return "latest"
+        else:
+            return PythonVersionSchema().dump(value)
+
+
+class VersionField(PythonVersionField):
+    def __init__(self, *args, **kwargs):
+        template = (
```
acroz: As with `VersionSchema` above.

pbugnion:
Fair enough -- I'll remove those classes.

PR #124, opened 2019-07-19
Add job schema and 'get job' method

snakch:
Added a get_job method and the relevant necessary schemas. 

Tried the method explicitly and it seems to work.

Might be worth checking if the ordering of schema and method definition makes sense


faculty/clients/job.py
```diff
@@ -125,6 +150,73 @@ def make_job_summary(self, data):
...
+    def make_job_command(self, data):
+        return JobCommand(**data)
+
+
+class JobDefinitionSchema(BaseSchema):
+    working_dir = fields.String(data_key="workingDir", required=True)
+    command = fields.Nested(JobCommandSchema, required=True)
+    image_type = fields.String(data_key="imageType", required=True)
+    conda_environment = fields.String(
+        data_key="condaEnvironment", missing=None, required=False
```
acroz: This should be either `missing=None` _or_ `required=False`.

tests/clients/test_job.py
```diff
@@ -84,7 +99,47 @@
...
+    "default": JOB_PARAMETERS.default,
+    "required": str(True),
+}
+JOB_COMMAND = JobCommand(name="command name", parameters=[JOB_PARAMETERS])
+JOB_COMMAND_BODY = {
+    "name": JOB_COMMAND.name,
+    "parameters": [JOB_PARAMETERS_BODY],
+}
+JOB_DEFINITION = JobDefinition(
+    working_dir="working dir",
```
acroz: A more representative example might be:
```suggestion
    working_dir="/project/subdir/",
```

tests/clients/test_job.py
```diff
@@ -84,7 +99,47 @@
...
+    command=JOB_COMMAND,
+    image_type="image type",
```
acroz: Same:
```suggestion
    image_type="m4.xlarge",
```

tests/clients/test_job.py
```diff
@@ -84,7 +99,47 @@
...
+    conda_environment="conda environment",
```
acroz: Same:
```suggestion
    conda_environment="Python3",
```

tests/clients/test_job.py
```diff
@@ -84,7 +99,47 @@
...
+    environment_ids=[str(ENVIRONMENT_ID)],
+    instance_size_type="instance size type",
```
acroz: As above:
```suggestion
    instance_size_type="m4.xlarge",
```
It's also the case that `instanceSizeType` must be `"custom"` if `instanceSize` is set. Should we have two tests - one with a 'custom' instance size and one with a preset like 'm4.xlarge' suggested above?

tests/clients/test_job.py
```diff
@@ -84,7 +99,47 @@
...
+)
+JOB_DEFINITION_BODY = {
+    "workingDir": JOB_DEFINITION.working_dir,
+    "command": JOB_COMMAND_BODY,
+    "imageType": JOB_DEFINITION.image_type,
+    "condaEnvironment": JOB_DEFINITION.conda_environment,
+    "environmentIds": [str(ENVIRONMENT_ID)],
+    "instanceSizeType": JOB_DEFINITION.instance_size_type,
+    "instanceSize": INSTANCE_SIZE_BODY,
+    "maxRuntimeSeconds": str(MAX_RUNTIME_SECONDS),
```
acroz: According to the Scala code this should be an integer.

tests/clients/test_job.py
```diff
@@ -84,7 +99,47 @@
 
 JOB_SUMMARY = JobSummary(id=JOB_ID, metadata=JOB_METADATA)
 JOB_SUMMARY_BODY = {"jobId": str(JOB_ID), "meta": JOB_METADATA_BODY}
-
+INSTANCE_SIZE = InstanceSize(milli_cpus=MILLI_CPUS, memory_mb=MEMORY_MB)
+INSTANCE_SIZE_BODY = {"milliCpus": str(MILLI_CPUS), "memoryMb": str(MEMORY_MB)}
```
acroz: According to the Scala code, these should be ints.

tests/clients/test_job.py
```diff
@@ -84,7 +99,47 @@
...
+JOB_PARAMETERS = JobParameter(
+    name="job name",
+    param_type=ParameterType.TEXT,
+    default="default parameter",
```
acroz: For clarity?:
```suggestion
    default="default parameter value",
```

tests/clients/test_job.py
```diff
@@ -84,7 +99,47 @@
...
+    required=True,
+)
+JOB_PARAMETERS_BODY = {
+    "name": JOB_PARAMETERS.name,
+    "type": "text",
+    "default": JOB_PARAMETERS.default,
+    "required": str(True),
```
acroz: JSON has boolean types; shouldn't this be just:
```suggestion
    "required": True,
```

tests/clients/test_job.py
```diff
@@ -84,7 +99,47 @@
...
+}
+JOB_COMMAND = JobCommand(name="command name", parameters=[JOB_PARAMETERS])
```
acroz: Unfortunately, the command 'name' is the actual command that gets run (it's a very poor choice of name that I tried, unsuccessfully, to change at the time). For clarity, could we do something like:
```suggestion
JOB_COMMAND = JobCommand(name="python myscript.py", parameters=[JOB_PARAMETERS])
```

faculty/clients/job.py
```diff
@@ -57,6 +62,26 @@ class RunState(Enum):
...
+        "command",
+        "image_type",
+        "conda_environment",
+        "environment_ids",
+        "instance_size_type",
+        "instance_size",
+        "max_runtime_seconds",
+    ],
+)
+Job = namedtuple("Job", ["meta", "definition"])
```
acroz: From looking at the Scala code, it looks like we also return the job ID with the job as the `jobId` key. Can we add it here as:
```suggestion
Job = namedtuple("Job", ["id", "meta", "definition"])
```
And then extract it in the schema with `data_key="jobId"`?

faculty/clients/job.py
```diff
@@ -125,6 +150,73 @@ def make_job_summary(self, data):
         return JobSummary(**data)
 
 
+class InstanceSizeSchema(BaseSchema):
+    milli_cpus = fields.Float(data_key="milliCpus", required=True)
```
acroz: These should be ints, not floats.

faculty/clients/job.py
```diff
@@ -125,6 +150,73 @@ def make_job_summary(self, data):
...
+    memory_mb = fields.Float(data_key="memoryMb", required=True)
+
+    @post_load
+    def make_instance_size(self, data):
+        return InstanceSize(**data)
+
+
+class JobParameterSchema(BaseSchema):
+    name = fields.String(required=True)
+    param_type = EnumField(
```
acroz: Why not just `'type`?

If it's because 'type' has special meaning in Python, I think it's ok to use it as an attribute as it's a builtin function, not a keyword, and as an attribute it won't conflict with the `type` builtin.

faculty/clients/job.py
```diff
@@ -125,6 +150,73 @@ def make_job_summary(self, data):
...
+
+    @post_load
+    def make_job_parameter(self, data):
+        return JobParameter(**data)
+
+
+class JobCommandSchema(BaseSchema):
+    name = fields.String(required=True)
+    parameters = fields.List(
+        fields.Nested(JobParameterSchema, required=True), required=True
```
acroz: I don't think this inner `required=True` has any effect since the parameter list can be empty.
```suggestion
        fields.Nested(JobParameterSchema), required=True
```

faculty/clients/job.py
```diff
@@ -125,6 +150,73 @@ def make_job_summary(self, data):
...
+    working_dir = fields.String(data_key="workingDir", required=True)
+    command = fields.Nested(JobCommandSchema, required=True)
+    image_type = fields.String(data_key="imageType", required=True)
+    conda_environment = fields.String(
+        data_key="condaEnvironment", missing=None, required=False
+    )
+    environment_ids = fields.List(
+        fields.String(), data_key="environmentIds", required=True
+    )
+    instance_size_type = fields.String(
```
acroz: What would you think about adding [schema level validation](https://marshmallow.readthedocs.io/en/3.0/extending.html#schema-level-validation) to check that the instance_size is set when the instance_size_type is 'custom' and not set otherwise?

I think it's important to then test both cases to catch (for example) the fact that the instance_size below is not always required.

faculty/clients/job.py
```diff
@@ -125,6 +150,73 @@ def make_job_summary(self, data):
...
+        data_key="instanceSizeType", required=True
+    )
+    instance_size = fields.Nested(
+        InstanceSizeSchema, data_key="instanceSize", required=False
```
acroz: As commented above.
```suggestion
        InstanceSizeSchema, data_key="instanceSize"
```

faculty/clients/job.py
```diff
@@ -140,7 +232,7 @@ class EnvironmentStepExecutionSchema(BaseSchema):
     ended_at = fields.DateTime(data_key="endedAt", missing=None)
 
     @post_load
-    def make_environment_step_execution_schema(self, data):
+    def make_environment_step_execution(self, data):
```
acroz: Good spot 👍

faculty/clients/job.py
```diff
@@ -256,6 +348,21 @@ def list(self, project_id):
         endpoint = "/project/{}/job".format(project_id)
         return self._get(endpoint, JobSummarySchema(many=True))
 
+    def get_job(self, project_id, job_id):
```
```suggestion
    def get(self, project_id, job_id):
```
To match `list` and the semantics of methods on other clients (methods for runs and subruns _do_ incude the name to distinguish them).

tests/clients/test_job.py
```diff
@@ -47,6 +58,10 @@
     SubrunSummarySchema,
 )
 
+MILLI_CPUS = 0.1
```
acroz: `MILLI_CPUS` should be an int -  a better example value would be 1000.

acroz: Could also just inline this and `MEMORY_MB` in `INSTANCE_SIZE` below as it's only used there.

Similar for `MAX_RUNTIME_SECONDS` in `JOB_DEFINITION`/

tests/clients/test_job.py
```diff
@@ -84,7 +99,47 @@
 
 JOB_SUMMARY = JobSummary(id=JOB_ID, metadata=JOB_METADATA)
 JOB_SUMMARY_BODY = {"jobId": str(JOB_ID), "meta": JOB_METADATA_BODY}
-
+INSTANCE_SIZE = InstanceSize(milli_cpus=MILLI_CPUS, memory_mb=MEMORY_MB)
+INSTANCE_SIZE_BODY = {"milliCpus": str(MILLI_CPUS), "memoryMb": str(MEMORY_MB)}
+JOB_PARAMETERS = JobParameter(
```
acroz: Since this is a single parameter:
```suggestion
JOB_PARAMETERS = JobParameter(
```

tests/clients/test_job.py
```diff
@@ -84,7 +99,47 @@
...
+    name="job name",
+    param_type=ParameterType.TEXT,
+    default="default parameter value",
+    required=True,
+)
+JOB_PARAMETERS_BODY = {
```
acroz: Ditto:
```suggestion
JOB_PARAMETER_BODY = {
```

tests/clients/test_job.py
```diff
@@ -210,6 +265,31 @@ def test_job_summary_schema():
...
+
+def test_job_definition_schema():
+    data = JobDefinitionSchema().load(JOB_DEFINITION_BODY)
+    assert data == JOB_DEFINITION
+
+
+def test_job_schema():
+    data = JobSchema().load(JOB_BODY)
+    assert data == JOB
+
```
acroz: Can we also add all of these schemas to the `test_schemas_invalid_data` parameterised test?

faculty/clients/job.py
```diff
@@ -194,13 +201,28 @@ class JobDefinitionSchema(BaseSchema):
...
+        ):
+            raise ValidationError(
+                "need to specify instance size for custom instances"
+            )
+        elif (
+            data["instance_size_type"] != "custom"
+            and data["instance_size"] is not None
+        ):
+            raise ValidationError(
+                "instance_size can only be specified if instance_size_type is 'custom' "
```
acroz: Looks like `flake8` is failing as this line is too long.

I suggest breaking the string across two lines, e.g.:

```python
            raise ValidationError(
                "instance_size can only be specified "
                "if instance_size_type is 'custom'"
            )
```

tests/clients/test_job.py
```diff
@@ -291,6 +292,11 @@ def test_job_definition_schema():
     assert data == JOB_DEFINITION
 
 
+def validate_job_definition_schema():
```
acroz: Functions need to start with `test_` to be detected as test by `pytest` - I think if you look at the logs when running the tests that this will not appear.

I think that our convention for a test of this sort would be to name it `test_job_definition_schema_invalid`.

It would be good if we could write a separate test for the other invalid case (when the instanceSizeType is `custom` and the `instanceSize` is unset. You could then write either a separate test for this case or try parameterising this existing test with `pytest.mark.parametrize` (see elsewhere in this file for an example).

faculty/clients/job.py
```diff
@@ -57,6 +67,26 @@ class RunState(Enum):
...
+        "command",
+        "image_type",
+        "conda_environment",
+        "environment_ids",
+        "instance_size_type",
+        "instance_size",
+        "max_runtime_seconds",
+    ],
+)
+Job = namedtuple("Job", ["job_id", "meta", "definition"])
```
acroz: Our convention (as you can see in `JobSummary` and the run / subrun namedtuples) is to have the primary UUID for an object be just `id` Can you rename the `job_id` attribute here to just `id` (and update the schemas and tests accordingly).

faculty/clients/job.py
```diff
@@ -125,6 +155,89 @@ def make_job_summary(self, data):
...
+
+    @post_load
+    def make_instance_size(self, data):
+        return InstanceSize(**data)
+
+
+class JobParameterSchema(BaseSchema):
+    name = fields.String(required=True)
+    type = EnumField(
+        ParameterType, data_key="type", by_value=True, required=True
```
acroz: Isn't `data_key="type"` redundant?

snakch:
I made the "create_job" branch a while back from the "job_schema" branch. I've addressed your comments on create_job - would it be ok for me to maybe rebase create_job to make things easier?

acroz:
It's going to be a pain to reproduce fixes that you've made on the other branch here, and I don't want to merge this branch with outstanding issues.

I think that the best thing to do will be to close this PR and make sure all these changes are present in your other PR.

In future, I'd suggest that if you're working on a second PR and notice some issue that was present in the first one, that you:

1. Check out the first branch
2. Make and commit the changes
3. Push the branch to update the first PR
4. Check out the second branch
5. Rebase on or merge the first branch to get the fix
6. Continue working on the second branch

In this case you'll have a first PR that can be reviewed and merged indepdenently, and be able to continue work on the second.

acroz:
@snakch and I have agreed to close this in preference to #125 as proposed above.

PR #125, opened 2019-07-19
Add client methods to get, create and update jobs

snakch:
added JobId Schema, create_job and relevant tests


faculty/clients/job.py
```diff
@@ -15,12 +15,27 @@
...
+    PYTHON = "python"
+    R = "r"
+
+
+class ParameterType(Enum):
+    TEXT = "text"
+    NUMBER = "number"
+
+
+class ImageType(Enum):
```
pbugnion: This class is duplicated

faculty/clients/job.py
```diff
@@ -256,9 +382,47 @@ def list(self, project_id):
...
         return self._get(endpoint, JobSummarySchema(many=True))
 
-    def update_metadata(self, project_id, job_id, name, description):
+    def create_job(self, project_id, job_parameters):
+        """Create a job.
+
+        Parameters
+        ----------
+        project_id : uuid.UUID
+        paramjob_parameters : Dict[dict]
```
pbugnion: this name is a little mangled.

tests/clients/test_job.py
```diff
@@ -84,7 +76,53 @@
...
+    working_dir="/project/subdir/",
+    command=JOB_COMMAND,
+    image_type=ImageType.PYTHON,
+    conda_environment="Python3",
+    environment_ids=[str(ENVIRONMENT_ID)],
+    instance_size_type="custom",
+    instance_size=INSTANCE_SIZE,
+    max_runtime_seconds=MAX_RUNTIME_SECONDS,
+)
+JOB_DEFINITION_BODY = {
```
acroz: As described in my comment on the implementation above, I think we need to also test the deserialisation of a non-custom image type job definition. Also, we should test the deserialisation of an R image type job definition (where the conda_environment will be `None`. As with the invalid test cases comment below, we probably want to make this a parametrized test.

snakch: Ok, I'll get to it - by parametrized test do you mean using the stuff described in here? http://doc.pytest.org/en/latest/example/parametrize.html

acroz: Confusingly, that's the 'advanced' parameterisation docs, I think you want more:
http://doc.pytest.org/en/latest/parametrize.html#parametrize-basics

Or just see the existing use of `pytest.mark.parametrize` in `test_job.py`.

faculty/clients/job.py
```diff
@@ -125,6 +154,93 @@ def make_job_summary(self, data):
...
+    conda_environment = fields.String(
+        data_key="condaEnvironment", missing=None
+    )
+    environment_ids = fields.List(
+        fields.String(), data_key="environmentIds", required=True
+    )
+    instance_size_type = fields.String(
+        data_key="instanceSizeType", required=True
+    )
+    instance_size = fields.Nested(InstanceSizeSchema, data_key="instanceSize")
```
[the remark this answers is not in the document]
snakch: Not sure I understand this comment - I've added tests to check that the instance size is set if and  only if instance_size_type is custom. Are you saying if we pass something like "m4.xlarge" as instace_size_type, deserialization will fail right now?

faculty/clients/job.py
```diff
@@ -125,6 +155,113 @@ def make_job_summary(self, data):
...
+    max_runtime_seconds = fields.Integer(
+        data_key="maxRuntimeSeconds", required=True
+    )
+
+    @validates_schema
+    def validate_conda_environment(self, data):
+        image_is_r = data["image_type"] == ImageType.R
+        print("HERE")
```
acroz: Need to remove these print statements.

faculty/clients/job.py
```diff
@@ -125,6 +155,113 @@ def make_job_summary(self, data):
...
+            print("IMAGE IS R")
+        image_is_python = data["image_type"] == ImageType.PYTHON
+        if image_is_python:
+            print("IMAGE IS PYTHON")
+        conda_environment_set = data["conda_environment"] is not None
+        if conda_environment_set:
+            print("CONDA ENVIRONMENT IS SET")
+        if image_is_r and conda_environment_set:
+            raise ValidationError(
+                "conda environment must only be set for python images"
```
acroz: Minor nit for consistency.
```suggestion
                "conda environment must only be set for Python images"
```

faculty/clients/job.py
snakch: oops!

tests/clients/test_job.py
```diff
@@ -210,6 +271,64 @@ def test_job_summary_schema():
...
+
+
+@pytest.mark.parametrize(
+    "instance_size_type, instance_size",
+    [("m4.xlarge", INSTANCE_SIZE), ("custom", None)],
+)
+def test_job_definition_schema_invalid_instance_type(
+    instance_size_type, instance_size
+):
+    JOB_DEFINITION_BODY_INVALID = JOB_DEFINITION_BODY.copy()
```
acroz: Since this is now being used as a local variable:
```suggestion
    job_definition_body_invalid = JOB_DEFINITION_BODY.copy()
```
or even just:
```suggestion
    invalid_body = JOB_DEFINITION_BODY.copy()
```

tests/clients/test_job.py
```diff
@@ -210,6 +271,64 @@ def test_job_summary_schema():
...
+
+
+@pytest.mark.parametrize(
+    "image_type, conda_environment",
+    [(ImageType.PYTHON, None), (ImageType.R, "Python3")],
+)
+def test_job_definition_schema_invalid_image_type(
+    image_type, conda_environment
+):
+    JOB_DEFINITION_BODY_INVALID = JOB_DEFINITION_BODY.copy()
```
acroz: As above (`JOB_DEFINITION_BODY_INVALID` -> `invalid_body` etc.)

tests/clients/test_job.py
```diff
@@ -210,6 +271,64 @@ def test_job_summary_schema():
...
+
+
+def test_job_schema():
+    data = JobSchema().load(JOB_BODY)
+    assert data == JOB
+
+
+def test_job_id_schema():
+    data = JobIdSchema().load({"jobId": str(JOB_ID)})
+    assert data == str(JOB_ID)
```
acroz: I think we want to get back a `uuid.UUID` object in this case, not a `str`.

faculty/clients/job.py
```diff
@@ -125,6 +155,113 @@ def make_job_summary(self, data):
...
+    meta = fields.Nested(JobMetadataSchema, required=True)
+    definition = fields.Nested(JobDefinitionSchema, required=True)
+
+    @post_load
+    def make_job(self, data):
+        return Job(**data)
+
+
+class JobIdSchema(BaseSchema):
+    jobId = fields.String(required=True)
```
```suggestion
    jobId = fields.UUID(required=True)
```

faculty/clients/job.py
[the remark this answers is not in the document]
acroz: Yes, I believe that is true.

The tests you've added to check these invalid cases are great - what I'm suggesting is to add an additional test for the other "happy path" where:

* `instance_size_type` is something other than custom
* `instance_size` is unset

If I'm right, this will currently fail, and we should then add `missing=None` to this line to fix.

faculty/clients/job.py
```diff
@@ -256,9 +387,47 @@ def list(self, project_id):
         endpoint = "/project/{}/job".format(project_id)
         return self._get(endpoint, JobSummarySchema(many=True))
 
-    def update_metadata(self, project_id, job_id, name, description):
+    def create_job(self, project_id, job_parameters):
```
acroz: For consistency with other methods, I think this should be called just `create`.

faculty/clients/job.py
```diff
@@ -256,9 +387,47 @@ def list(self, project_id):
...
+        """Create a job.
+
+        Parameters
+        ----------
+        project_id : uuid.UUID
+        job_parameters : Dict[dict]
```
acroz: Users of this API will have to know how to construct a fairly complicated nested dictionary. Can we instead have an API like:

```python
def create(
    self,
    project_id,
    name,
    description,
    definition
):
    # etc.
```

Then the user can pass in a `JobDefinition` object, which is harder to get wrong (though admittedly still non-trivial).

Note that I've deliberately not used the `JobMetadata` object as an argument, because:
* We can't set the `author_id`, `created_at` or `last_updated_at` so would need a new version of that object with those attributes omitted.
* There are only two keys remaining, and it seems a bit over the top to force the user to put them in an object first.
* It also mirrors the signature of `update_metadata`.

We can then use something like the following to create the payload:

```python
schema = JobDefinitionSchema()
payload = {
    "meta": {
        "name": name,
        "description": description
    },
    "definition": schema.dump(definition)
}
```

faculty/clients/job.py
```diff
@@ -256,9 +387,47 @@ def list(self, project_id):
...
+            A dictionnary containing the metadata and definition of the job to
+            be created.
+
+        Returns
+        -------
+        string
```
```suggestion
        uuid.UUID
```

faculty/clients/job.py
```diff
@@ -256,9 +387,46 @@ def list(self, project_id):
...
-    def update_metadata(self, project_id, job_id, name, description):
+    def create(self, project_id, name, description, definition):
+        """Create a job.
+
+        Parameters
+        ----------
+        project_id : uuid.UUID
+        name: str
+        description: str
+        definition: Dict
```
acroz: I more had in mind that users would pass a `JobDefinition` object, which we would then convert to the corresponding `dict` with the `JobDefinitionSchema`'s `dump` method. What do you think?

acroz:
@snakch will do a final re-test against an actual platform deployment before we merge.

PR #128, opened 2019-07-25
Refactor datasets to use new object store service and presigned URLs

eliasbenussi:


faculty/clients/object.py
```diff
@@ -153,6 +186,32 @@ def list(self, project_id, prefix="/", page_token=None):
...
+        PathAlreadyExists
+            when the path that we want to create as a directory already exists
+        """
+        endpoint = "/project/{}/directory/{}".format(
+            project_id, path.lstrip("/")
+        )
+        try:
+            self._put_raw(endpoint)
+        except Conflict as err:
+            if err.error_code == "object_already_exists":
```
mociarain: What would you think about adding the error_code to the relevant "error" class as a class variable since there's a one-to-one mapping and using it here?

acroz: In that case we'd probably also want to embed the HTTP error exception class expected as well (i.e. `Conflict`). While the current implementation is not very clever, it is quite readable and I suggest keeping it.

faculty/clients/object.py
```diff
@@ -153,6 +186,32 @@ def list(self, project_id, prefix="/", page_token=None):
             params["pageToken"] = page_token
         return self._get(endpoint, ListObjectsResponseSchema(), params=params)
 
+    def create_directory(self, project_id, path):
+        """Copy objects in the store.
```
acroz: Incorrect description.

faculty/datasets/__init__.py
```diff
@@ -160,26 +125,27 @@ def _isdir(project_path, project_id=None, s3_client=None):
...
+    object_client : faculty.clients.object.ObjectClient, optional
+        Advanced - can be used to benefit from caching in chain interactions
+        with datasets.
 
     Returns
     -------
     bool
     """
-    # 'Directories' in the S3 bucket always end in a '/'
+    # 'Directories' always end in a '/' in S3
```
acroz: Remove S3 reference?

tests/datasets/test_datasets.py
```diff
@@ -13,363 +13,433 @@
 # limitations under the License.
 
 
-import fnmatch
-import os
-import posixpath
+# import fnmatch
+# import os
+# import posixpath
```
acroz: Unused?

tests/datasets/test_datasets.py
```diff
@@ -13,363 +13,433 @@
...
 import pytest
 
+# import mock
```
acroz: Unused?

tests/datasets/test_datasets.py
```diff
@@ -13,363 +13,433 @@
...
+#     TEST_TREE,
+#     TEST_TREE_NO_HIDDEN_FILES,
+#     VALID_ROOT_PATHS,
+#     VALID_DIRECTORIES,
+#     VALID_FILES,
+#     INVALID_PATHS,
+#     TEST_FILE_CONTENT,
+#     TEST_NON_EXISTENT,
+#     EMPTY_DIRECTORY,
+# )
```
acroz: Unused?

tests/datasets/test_datasets.py
```diff
@@ -13,363 +13,433 @@
...
+    get_session_mock = mocker.patch(
+        "faculty.datasets.get_session", return_value=session
+    )
+
+    project_id = mocker.Mock()
+
+    object_client = mocker.Mock()
+    mocker.patch("faculty.datasets.ObjectClient", return_value=object_client)
+
+    yield [object_client, project_id]
```
acroz: Nit: I think this would make more sense as a tuple:
```suggestion
    yield (object_client, project_id)
```
or even just:
```suggestion
    yield object_client, project_id
```

tests/datasets/test_datasets.py
```diff
@@ -13,363 +13,433 @@
...
-    [("/input/no-such-file", "/output/new-file"), ("/input/", "/output/")],
-)
-def test_cp_directory_source(remote_dir, destination):
-    with pytest.raises(datasets.DatasetsError):
-        datasets.cp(remote_dir, destination)
+    object_client.create_directory.return_value = mocker.Mock()
 
+    entry_mock = "test-file"
+    os_listdir_mock = mocker.patch(
+        "faculty.datasets.os.listdir", return_value=[entry_mock]
```
acroz:
```suggestion
        "os.listdir", return_value=[entry_mock]
```

faculty/datasets/__init__.py
```diff
@@ -20,38 +20,18 @@
...
-from faculty.datasets.session import DatasetsError
+from faculty.session import get_session
+from faculty.context import get_context
+from faculty.clients.object import ObjectClient
+from faculty.datasets import path, transfer
+from faculty.datasets.path import DatasetsError
 
 
 # For backwards compatibility
 SherlockMLDatasetsError = DatasetsError
```
acroz: Should we take the opportunity to remove this?

faculty/datasets/__init__.py
```diff
@@ -190,94 +156,58 @@ def _isfile(project_path, project_id=None, s3_client=None):
...
-        msg = (
-            "the destination path {} indicates a directory but the "
-            "source path {} is a normal file - please provide a full "
-            "destination path"
-        ).format(repr(project_path), repr(local_path))
-        raise DatasetsError(msg)
-
-    s3_client.upload_file(local_path, bucket, bucket_path)
+def _create_parent_directories(project_path, project_id, object_client):
+    parent_path = path.project_parent_directory(project_path)
```
acroz: I suggest we just always do:

```python
parent_path = posixpath.dirname(project_path)
object_client.create_directory(project_id, parent_path, parents=True)
```

This should just work under all circumstances, and we can drop `path.project_parent_directory`.

faculty/datasets/path.py
```diff
@@ -93,3 +78,30 @@ def project_parent_directories(project_path):
         directories.append("/".join(parts[:i_last]) + "/")
 
     return directories
+
+
+def project_parent_directory(project_path):
```
acroz: GitHub won't let me comment there, but it seems we're not using `project_parent_directories` (above) any more?

acroz: Also see earlier comment that may imply removing this function too.

tests/clients/test_object.py
```diff
@@ -227,6 +230,33 @@ def test_object_client_list_defaults(mocker):
...
 
+def test_object_client_create_directory_default(mocker):
+    mocker.patch.object(ObjectClient, "_put_raw")
+
+    client = ObjectClient(mocker.Mock())
+    client.create_directory(PROJECT_ID, "test-path")
+
+    ObjectClient._put_raw.assert_called_once_with(
+        "/project/{}/directory/{}".format(PROJECT_ID, "test-path"),
+        params={"parents": 0},
```
acroz: Can we test with `parents=True` (perhaps with a parameterised test)?

tests/clients/test_object.py
```diff
@@ -263,7 +293,7 @@ def test_object_client_copy_multiple(mocker):
     )
```
acroz: Maybe change the name of the above test to `test_object_client_copy_recursive`, or merge it and `test_object_client_copy_default` into a single parameterised test?

tests/clients/test_object.py
```diff
@@ -281,7 +321,8 @@ def test_object_client_delete(mocker):
     client.delete(PROJECT_ID, path)
 
     ObjectClient._delete_raw.assert_called_once_with(
-        "/project/{}/object/{}".format(PROJECT_ID, path), params={}
+        "/project/{}/object/{}".format(PROJECT_ID, path),
+        params={"recursive": 0},
```
acroz: Can we test the recursive case, perhaps with a parameterised test?

tests/datasets/test__init__.py
```diff
@@ -0,0 +1,424 @@
+# Copyright 2018-2019 Faculty Science Limited
```
acroz: We would normally call a file like this `test_init.py` (note just single underscore).

tests/datasets/test__init__.py
```diff
@@ -0,0 +1,424 @@
...
+
+
+@pytest.fixture
+def mock_client(mocker):
+    session = mocker.Mock()
+    get_session_mock = mocker.patch(
+        "faculty.datasets.get_session", return_value=session
+    )
+
+    project_id = mocker.Mock()
```
acroz: It seems a bit odd that this is created by the `mock_client` fixture, which otherwise does not seem to depend on project ID.

We could alternatively just have a single `uuid.UUID` in the module namespace used by all tests:

```
import uuid

PROJECT_ID = uuid.uuid4()
```

tests/datasets/test__init__.py
```diff
@@ -0,0 +1,424 @@
...
+    object_client.list.assert_called_once_with(project_id, "test-prefix")
+
+
+def test_datasets_ls_non_hidden_files(mocker, mock_client):
+    object_client, project_id = mock_client
+
+    list_response = mocker.Mock()
+    mock_object = mocker.Mock()
+    mock_object.path = ".test-hidden-path"
+    list_response.objects = [mock_object]
```
acroz: Might be good to also check that normal files are not filtered out. In this situation, the following more concise way of building `Mock` objects might be useful:

```python
list_response.objects = [mocker.Mock(path=".test-hidden"), mocker.Mock(path="not-hidden")]
```

tests/datasets/test__init__.py
```diff
@@ -0,0 +1,424 @@
...
+    )
+
+    pattern_mock = mocker.Mock()
+    datasets.glob(
+        pattern_mock,
+        prefix="project-path",
+        project_id=project_id,
+        show_hidden=True,
+        object_client=None,
+    )
```
acroz: Want to assert that this returns the return value of `fnmatch.filter`?

I'd also be happy not mocking `fnmatch.filter` in this case.

tests/datasets/test__init__.py
```diff
@@ -0,0 +1,424 @@
...
+                object_client=object_client,
+            ),
+        ]
+    )
+    _get_file_mock.assert_called_once_with(
+        "project-path/test-file", local_dests[1], project_id, object_client
+    )
+
+
+def test_datasets_put_file_and_create_parent_directories(mocker, mock_client):
```
acroz:
```suggestion
def test_datasets_put_file(mocker, mock_client):
```

tests/datasets/test__init__.py
```diff
@@ -0,0 +1,424 @@
...
+        show_hidden=True,
+        object_client=None,
+    )
+    fnmatch_filter_mock.assert_called_once_with(content_mock, pattern_mock)
+
+
+def test_datasets_get_file(mocker, mock_client):
+    object_client, project_id = mock_client
+
+    ls_mock = mocker.patch("faculty.datasets.ls", return_value=[])
```
acroz: Would this (and subsequent tests) be made simpler by testing `_isdir` separately and then mocking it here?

tests/datasets/test__init__.py
```diff
@@ -0,0 +1,424 @@
...
+    os_listdir_mock.assert_called_once_with("local-path")
+    _put_file_mock.assert_called_once_with(
+        "local-path/test-file",
+        "project-path/test-file",
+        project_id,
+        object_client,
+    )
+
+
+def test_datasets_cp_path(mocker, mock_client):
```
acroz:
```suggestion
def test_datasets_cp(mocker, mock_client):
```

tests/datasets/test__init__.py
```diff
@@ -0,0 +1,424 @@
...
+    object_client, project_id = mock_client
+    object_client.copy.return_value = mocker.Mock()
+
+    path_project_parent_directory = mocker.patch(
+        "faculty.datasets.path.project_parent_directory", return_value="/"
+    )
+    object_client.create_directory.return_value = mocker.Mock()
```
acroz: This is essentially a noop. The return value will be a `Mock` object by default.

tests/datasets/test__init__.py
```diff
@@ -0,0 +1,424 @@
...
+        project_id, "/", parents=True
+    )
+    object_client.copy.assert_called_once_with(
+        project_id, "source-path", "destination-path", recursive=True
+    )
+
+
+def test_datasets_rm_path(mocker, mock_client):
+    object_client, project_id = mock_client
+    object_client.delete.return_value = mocker.Mock()
```
acroz: This is essentially a noop. The return value will be a Mock object by default.

tests/datasets/test__init__.py
```diff
@@ -0,0 +1,424 @@
...
+    path_project_parent_directory.assert_called_once_with("destination-path")
+    object_client.create_directory.assert_called_once_with(
+        project_id, "/", parents=True
+    )
+    object_client.copy.assert_called_once_with(
+        project_id, "source-path", "destination-path", recursive=True
+    )
+
+
+def test_datasets_rm_path(mocker, mock_client):
```
acroz:
```suggestion
def test_datasets_rm(mocker, mock_client):
```

tests/datasets/test__init__.py
```diff
@@ -0,0 +1,424 @@
...
+    object_client.delete.return_value = mocker.Mock()
+
+    datasets.rm("project-path", project_id=project_id, recursive=True)
+
+    object_client.delete.assert_called_once_with(
+        project_id, "project-path", recursive=True
+    )
+
+
+def test_datasets_rmdir_path(mocker):
```
acroz:
```suggestion
def test_datasets_rmdir(mocker):
```

tests/datasets/test__init__.py
```diff
@@ -0,0 +1,424 @@
...
+    project_id = mocker.Mock()
+
+    rm_mock = mocker.patch("faculty.datasets.rm", return_value=mocker.Mock())
```
acroz: `return_value=mocker.Mock()` is the default.
```suggestion
    rm_mock = mocker.patch("faculty.datasets.rm")
```

tests/datasets/test__init__.py
```diff
@@ -0,0 +1,424 @@
...
+        project_id=project_id,
+        recursive=True,
+        object_client=None,
+    )
+
+
+def test_datasets_mv_path(mocker, mock_client):
+    object_client, project_id = mock_client
+
+    cp_mock = mocker.patch("faculty.datasets.cp", return_value=mocker.Mock())
```
acroz: As above
```suggestion
    cp_mock = mocker.patch("faculty.datasets.cp")
```

tests/datasets/test__init__.py
```diff
@@ -0,0 +1,424 @@
...
+    rm_mock = mocker.patch("faculty.datasets.rm", return_value=mocker.Mock())
```
acroz:
```suggestion
    rm_mock = mocker.patch("faculty.datasets.rm")
```

tox.ini
```diff
@@ -9,6 +9,7 @@ deps =
     pytest-mock
     requests_mock
     python-dateutil>=2.7
+    mock
```
acroz: Where are we using this?

tox.ini
```diff
@@ -9,6 +9,7 @@ deps =
     pytest-mock
```
acroz: GitHub won't let me comment directly, but we can probably remove the `passenv` line above that was there to pass through credentials for the integration tests.

faculty/datasets/__init__.py
```diff
@@ -454,23 +396,22 @@ def rm(project_path, project_id=None, s3_client=None):
...
 
-    bucket = _bucket(project_id)
-    bucket_path = path.projectpath_to_bucketpath(project_path, project_id)
+    object_client.delete(project_id, project_path, recursive=recursive)
 
-    s3_client.delete_object(Bucket=bucket, Key=bucket_path)
 
-
-def rmdir(project_path, project_id=None):
+def rmdir(project_path, project_id=None, object_client=None):
```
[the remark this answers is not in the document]
eliasbenussi: To keep things as similar to the old API as possible, I opted for only allowing it on empty directories

faculty/datasets/__init__.py
```diff
@@ -515,20 +446,21 @@ def etag(project_path, project_id=None):
...
-    bucket = _bucket(project_id)
-    bucket_path = path.projectpath_to_bucketpath(project_path, project_id)
+    project_id = project_id or get_context().project_id
+    object_client = object_client or ObjectClient(get_session())
 
-    s3_object = client.get_object(Bucket=bucket, Key=bucket_path)
+    object = object_client.get(project_id, project_path)
 
-    return s3_object["ETag"].strip('"')
+    return object.etag.strip('"')
```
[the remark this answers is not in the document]
eliasbenussi: That's a good point. It does, I double checked. I will remove the strip from here and do it in Hoard

faculty/datasets/__init__.py
eliasbenussi: @acroz and I spoke and decided not to remove this yet

tests/clients/test_object.py
```diff
@@ -227,43 +230,74 @@ def test_object_client_list_defaults(mocker):
...
 
     ObjectClient._put_raw.assert_called_once_with(
-        "/project/{}/object/{}".format(PROJECT_ID, "destination"),
-        params={"sourcePath": "source"},
+        "/project/{}/directory/{}".format(PROJECT_ID, "test-path"),
+        params={"parents": 0},
+    )
+
+
+@pytest.mark.parametrize("parents,expected_parent", [(True, 1), (False, 0)])
```
acroz:
```suggestion
@pytest.mark.parametrize("parents, expected_parent", [(True, 1), (False, 0)])
```

tests/clients/test_object.py
```diff
@@ -227,43 +230,74 @@ def test_object_client_list_defaults(mocker):
...
 
     ObjectClient._put_raw.assert_called_once_with(
         "/project/{}/object/{}".format(PROJECT_ID, "destination"),
         params={"sourcePath": "source", "recursive": 0},
     )
 
 
-def test_object_client_copy_multiple(mocker):
+@pytest.mark.parametrize(
+    "recursive,expected_recursive", [(True, 1), (False, 0)]
```
acroz:
```suggestion
    "recursive, expected_recursive", [(True, 1), (False, 0)]
```

tests/clients/test_object.py
```diff
@@ -273,15 +307,42 @@ def test_object_client_copy_single_source_not_found(mocker):
...
 
     ObjectClient._delete_raw.assert_called_once_with(
-        "/project/{}/object/{}".format(PROJECT_ID, path), params={}
+        "/project/{}/object/{}".format(PROJECT_ID, path),
+        params={"recursive": 0},
+    )
+
+
+@pytest.mark.parametrize(
+    "recursive,expected_recursive", [(True, 1), (False, 0)]
```
acroz:
```suggestion
    "recursive, expected_recursive", [(True, 1), (False, 0)]
```

tests/datasets/test_init.py
```diff
@@ -0,0 +1,402 @@
...
+        project_id=PROJECT_ID,
+        show_hidden=True,
+        object_client=None,
+    )
+    rm_mock.assert_called_once_with(
+        "project-path", project_id=PROJECT_ID, object_client=None
+    )
+
+
+def test_datasets_rmdir_empty_directory_raises(mocker):
```
acroz:
```suggestion
def test_datasets_rmdir_nonempty_directory_raises(mocker):
```

faculty/datasets/__init__.py
```diff
@@ -481,30 +421,24 @@ def rmdir(project_path, project_id=None):
...
     """
 
-    s3_client = _s3_client(project_id)
-
-    if not _isdir(project_path, project_id, s3_client):
-        raise DatasetsError("not a directory")
-
     contents = ls(
-        project_path, project_id, show_hidden=True, s3_client=s3_client
+        prefix=project_path,
```
acroz: What if we list with project_path `/path/to/dir` or `/path/to/dir/` and get back the directory as an object?

acroz: Also, I believe this method will allow you to delete files, which would be wrong.

PR #129, opened 2019-07-30
Implement GCS upload in datasets

mociarain:
This PR contains the functionality to upload files and streams to GCS

### DIFF
--- a/tests/datasets/test_transfer.py
+++ b/tests/datasets/test_transfer.py
@@ -27,7 +27,7 @@ PROJECT_ID = uuid4()
 TEST_PATH = "/path/to/file"
 TEST_URL = "https://example.com/presigned/url"
 OTHER_URL = "https://example.com/other-presigned/url"
-TEST_AWS_UPLOAD_ID = 123
+TEST_S3_UPLOAD_ID = 123
 
 TEST_ETAG = "5d24e152bcdfa5a0357f46471be3be6c"
 TEST_COMPLETED_PART = CompletedUploadPart(1, TEST_ETAG)
@@ -58,12 +58,12 @@ def mock_client_download(mocker, requests_mock):
 def mock_client_upload_s3(mocker, requests_mock):
     presigned_response_mock = mocker.Mock()
     presigned_response_mock.provider = CloudStorageProvider.S3
-    presigned_response_mock.upload_id = TEST_AWS_UPLOAD_ID
+    presigned_response_mock.upload_id = TEST_S3_UPLOAD_ID
 
     object_client = mocker.Mock()
     object_client.presign_upload.return_value = presigned_response_mock
 
-    yield object_client
+    return object_client
 
 
 @pytest.fixture
@@ -75,7 +75,7 @@ def mock_client_upload_gcs(mocker, requests_mock):
     object_client = mocker.Mock()
     object_client.presign_upload.return_value = presigned_response_mock
 
-    yield object_client
+    return object_client
 
 
 def _assert_contains(dict_1, dict_2):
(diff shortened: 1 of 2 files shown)


acroz:
@mociarain This can now be rebased off master.

acroz:
@mociarain I've done the rebase myself for the sake of saving time.

tests/datasets/test_transfer.py
```diff
@@ -0,0 +1,229 @@
...
+
+from faculty.clients.object import CloudStorageProvider, CompletedUploadPart
+from faculty.datasets import transfer
+
+
+PROJECT_ID = uuid4()
+TEST_PATH = "/path/to/file"
+TEST_URL = "https://example.com/presigned/url"
+OTHER_URL = "https://example.com/other-presigned/url"
+TEST_AWS_UPLOAD_ID = 123
```
acroz: S3 upload IDs are strings.

Can we also:
```suggestion
TEST_S3_UPLOAD_ID = 123
```

tests/datasets/test_transfer.py
```diff
@@ -0,0 +1,229 @@
...
+@pytest.fixture
+def mock_client_upload_s3(mocker, requests_mock):
+    presigned_response_mock = mocker.Mock()
+    presigned_response_mock.provider = CloudStorageProvider.S3
+    presigned_response_mock.upload_id = TEST_AWS_UPLOAD_ID
+
+    object_client = mocker.Mock()
+    object_client.presign_upload.return_value = presigned_response_mock
+
+    yield object_client
```
acroz: If we don't want this fixture to do anything after the test finishes, it does not need to be a generator and we can just:
```suggestion
    return object_client
```
However, we may want to keep it as a generator and add the following:
```python
object_client.presign_upload.assert_called_once_with(PROJECT_ID, TEST_PATH)
```

tests/datasets/test_transfer.py
```diff
@@ -0,0 +1,229 @@
...
+@pytest.fixture
+def mock_client_upload_gcs(mocker, requests_mock):
+    presigned_response_mock = mocker.Mock()
+    presigned_response_mock.provider = CloudStorageProvider.GCS
+    presigned_response_mock.url = TEST_URL
+
+    object_client = mocker.Mock()
+    object_client.presign_upload.return_value = presigned_response_mock
+
+    yield object_client
```
acroz: As above.

tests/datasets/test_transfer.py
```diff
@@ -0,0 +1,229 @@
...
+    transfer.download_file(
+        mock_client_download, PROJECT_ID, TEST_PATH, destination
+    )
+    assert destination.read(mode="rb") == TEST_CONTENT
+
+
+def test_chunking_and_labelling_data_of_exact_sizes(mocker):
+    mocker.patch("faculty.datasets.transfer.UPLOAD_CHUNK_SIZE", 4)
+    content = [b"1111", b"2222", b"last"]
+    a = transfer._rechunk_and_label_as_last(content)
```
acroz: "a" isn't a great variable name. Can we change it to something like "chunks"?

tests/datasets/test_transfer.py
```diff
@@ -0,0 +1,229 @@
...
+    assert list(a) == [(b"1111", False), (b"2222", False), (b"last", True)]
+
+
+def test_chunking_and_labelling_of_smaller_sizes(mocker):
+    mocker.patch("faculty.datasets.transfer.UPLOAD_CHUNK_SIZE", 12)
+    content = [b"1111", b"2222", b"last"]
+    a = transfer._rechunk_and_label_as_last(content)
```
acroz: As above:
"a" isn't a great variable name. Can we change it to something like "chunks"?

tests/datasets/test_transfer.py
```diff
@@ -0,0 +1,229 @@
...
+    assert list(a) == [(b"11112222last", True)]
+
+
+def test_chunking_and_labelling_of_greater_sizes(mocker):
+    mocker.patch("faculty.datasets.transfer.UPLOAD_CHUNK_SIZE", 4)
+    content = [b"11112222last"]
+    a = transfer._rechunk_and_label_as_last(content)
```
acroz: As above:
"a" isn't a great variable name. Can we change it to something like "chunks"?

tests/datasets/test_transfer.py
mociarain: I think I just forgot the assertion. I'll add that now

PR #133, opened 2019-08-21
Yield an empty chunk

mociarain:


tests/datasets/test_transfer.py
```diff
@@ -108,6 +108,15 @@ def test_download_file(mock_client_download, tmpdir):
     assert destination.read(mode="rb") == TEST_CONTENT
 
 
+@pytest.mark.parametrize(
+    "content, expected_chunks", [([], [b""]), ([b""], [b""])]
+)
+def test_chunking_of_empty_files(mocker, content, expected_chunks):
```
acroz: `expected_chunks` is always `[b""]` - probably would be more readable to not parametrize it.

PR #137, opened 2019-09-05
Compress multipart upload request

acroz:
This is a low-effort solution to help the request get through proxies with a low max body size.

In addition, we'll increase the max body size for this service on platform deployments.

### DIFF
--- a/faculty/clients/object.py
+++ b/faculty/clients/object.py
@@ -15,6 +15,8 @@
 
 from collections import namedtuple
 from enum import Enum
+import json
+import gzip
 
 from marshmallow import fields, post_load
 from marshmallow_enum import EnumField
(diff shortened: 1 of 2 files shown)


srstevenson:
Travis is failing because [`gzip.compress`](https://docs.python.org/3/library/gzip.html#gzip.compress) isn't available in Python 2.7.

acroz:
@zblz With your config changes on the backend of this endpoint, is this change redundant? From what I understood, the limit applies to the uncompressed payload size.

zblz:
It will help with the limits on the nginx proxies and general transfer speed (specially on slow connections), but it will indeed not help with the limit we were hitting up to now. Given the hoops we might have to jump for Python 2 and 3 support, I'm on the fence about whether it's worth keeping it.

zblz:
We agreed offline that this is not worth it given the need to implement different Python 2 and 3 code flows.

PR #138, opened 2019-09-09
Url encode object paths

mociarain:
This addresses a bug Liam noticed in datasets where path string are placed directly in url's unencoded


liamcoatman:
@mociarain can confirm that this fixes the bug I had

PR #139, opened 2019-09-30
Don't sent context-range uploading empty file

eliasbenussi:


faculty/datasets/transfer.py
```diff
@@ -209,17 +209,14 @@ def _gcs_upload(upload_url, content):
...
-        data=content,
-        headers={
-            "Content-Length": "{0}".format(len(content)),
-            "Content-Range": "bytes {0}-{1}/{2}".format(
-                start_index, end_index, total_file_size
-            ),
-        },
-    )
+    headers = {"Content-Length": "{0}".format(len(content))}
+    if start_index or content:
```
acroz: Can we add a comment to the code to explain the purpose of this condition?

acroz:
LGTM - good find. Suggested adding a comment but apart from that happy to merge and release.

acroz:
Merged this as I can push it through with other changes.

Issue #130, opened 2019-07-31
Environment creation without description


liamcoatman:
I create a new environment with `description=None` (the default). When I then call the `list` method on the environment client I get a marshmallow validation error because the type of the `description` field in the `EnvironmentSchema` is `fields.String`. Adding `allow_none=True` to the `description` field fixes the problem (this is used in `EnvironmentCreateUpdateSchema`). However, environments created using the client without descriptions have `description=None` whereas environments created through the UI have `description=''`.

mociarain:
Fixed by: https://github.com/facultyai/faculty/pull/142

PR #142, opened 2019-11-13
Fix environment model

mociarain:
Currently if you list environments for a project and one of them has an empty description the command fails with:
`marshmallow.exceptions.ValidationError: {1: {'description': ['Field may not be null.']}}`

This is incorrect as the description for an environment can be None (see EnvironmentCreateUpdateSchema).  This PR fixes this and adds a test to prevent regression. I've also tested it successfully against one of the dev clusters

### DIFF
--- a/faculty/clients/environment.py
+++ b/faculty/clients/environment.py
@@ -233,7 +233,7 @@ class EnvironmentSchema(BaseSchema):
     id = fields.UUID(data_key="environmentId", required=True)
     project_id = fields.UUID(data_key="projectId", required=True)
     name = fields.String(required=True)
-    description = fields.String(required=True)
+    description = fields.String(required=True, allow_none=True)
     author_id = fields.UUID(data_key="authorId", required=True)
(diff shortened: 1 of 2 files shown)


faculty/clients/environment.py
```diff
@@ -233,7 +233,7 @@ class EnvironmentSchema(BaseSchema):
     id = fields.UUID(data_key="environmentId", required=True)
     project_id = fields.UUID(data_key="projectId", required=True)
     name = fields.String(required=True)
-    description = fields.String(required=True)
+    description = fields.String(required=True, allow_none=True)
```
acroz: We usually treat 'None' as equivalent to missing - can we make this `missing=None` instead of `required=True, allow_none=True`?

mociarain: Agreed. That's much clearer. I'll change it in the EnvironmentCreateUpdateSchema too for consistency

liamcoatman:
I think this fixes #130

PR #144, opened 2019-12-12
Make arguments to faculty.client explicit and add docstring

acroz:
It's currently hard to understand which arguments you can pass to `faculty.client`. This PR makes them explicit instead of relying on `**kwargs` and adds a docstring.

### DIFF
--- a/faculty/__init__.py
+++ b/faculty/__init__.py
@@ -75,9 +75,9 @@ def client(
     ... )
     <faculty.clients.account.AccountClient object at 0x10e447550>
 
-    To override any of the domain, protocol, client ID or client secret read
-    from a credentials file or from environment variables, pass their
-    respective arguments:
+    To set any of the domain, protocol, client ID or client secret, pass their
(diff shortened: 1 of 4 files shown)


faculty/__init__.py
```diff
@@ -17,7 +17,85 @@
...
+
+    >>> faculty.client(
+    ...     "account",
+    ...     domain="services.faculty.myorganisation.com",
+    ...     protocol="http",
+    ...     client_id="d047dad1-175b-4810-84b1-c0ff6bbcf456",
+    ...     client_secret="Vk4a1FgSKlTplkK0GkToAdRTdsoU4XHuwCMQ",
+    ... )
+    <faculty.clients.account.AccountClient object at 0x10e4472b0>
+    """
```
zblz: Should we mention what happens if you pass both `credentials_path/profile_name` *and* `domain/protocol/client_id/client_secret`?

acroz: I had covered this, but it wasn't super clear. Have reworded to clarify.

PR #147, opened 2019-12-18
Chunk sizing based on max chunks for S3 + tests

jkeelan:
Allows for files over 50gb to uploaded to S3. Dynamically sets chunk size to be ceil (size of file / 10000) if file is greater than 50 gb. GCS remains at 5MB chunk.

### DIFF
--- a/faculty/datasets/transfer.py
+++ b/faculty/datasets/transfer.py
@@ -139,8 +139,6 @@ def upload_stream(object_client, project_id, datasets_path, content):
         The target path to upload to in the object store
     content : Iterable[bytes]
         The data to upload, chunked
-    chunk_size : int
-        The size of chunk, in bytes, to use during upload
     """
     _upload_stream(object_client, project_id, datasets_path, content)
 
--- a/faculty/datasets/transfer.py
+++ b/faculty/datasets/transfer.py
@@ -119,7 +119,6 @@ def upload(object_client, project_id, datasets_path, content):
         The data to upload
     """
     # upload_stream will rechunk the data anyway so just pass as a single chunk
-
     _upload_stream(
         object_client,
         project_id,
--- a/faculty/datasets/transfer.py
+++ b/faculty/datasets/transfer.py
@@ -170,13 +170,9 @@ def _upload_stream(
 ):
 
     presign_response = object_client.presign_upload(project_id, datasets_path)
+    chunk_size = _chunk_size(presign_response.provider, known_file_size)
 
     if presign_response.provider == CloudStorageProvider.S3:
-        if known_file_size is None:
-            chunk_size = DEFAULT_CHUNK_SIZE
-        else:
-            chunk_size = _s3_chunk_size(known_file_size)
-
         _s3_upload(
             object_client,
             project_id,
@@ -186,11 +182,6 @@ def _upload_stream(
             chunk_size,
         )
     elif presign_response.provider == CloudStorageProvider.GCS:
-        if known_file_size is None:
-            chunk_size = DEFAULT_CHUNK_SIZE
-        else:
-            chunk_size = _gcs_chunk_size(known_file_size)
-
         _gcs_upload(presign_response.url, content, chunk_size)
     else:
         raise ValueError(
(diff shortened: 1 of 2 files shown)


pbugnion:
Thanks for submitting this!

Unless I'm missing something, it looks like this iterates through the content to get the size. This will presumably be absurdly slow on a file of any size. By contrast, I believe a lookup on the inode table for the file size (e.g. using tooling in `os`) will be independent of the size (?). At least, it will be much quicker.

It might be worth looking at the source code for other tools that do this (`awscli`, `s4cmd`, `s3cmd`) to see how they calculate the chunk size.

jkeelan:
Based on the feedback I've moved the upload logic to the private _upload_stream function which takes an optional known_file_size argument. Of the three public functions, upload_file and upload provide this argument, which then calculates a new chunk size for S3. The new chunk size is used only if it exceeds the DEFAULT_CHUNK_SIZE.

upload_stream just uses the DEFAULT_CHUNK_SIZE. In addition, the initial file chunking uses a separate chunk size FILE_CHUNK_SIZE.

faculty/datasets/transfer.py
```diff
@@ -160,19 +155,57 @@ def upload_file(object_client, project_id, datasets_path, local_path):
...
+def _upload_stream(
+    object_client, project_id, datasets_path, content, known_file_size=None
+):
 
-    completed_parts = []
+    presign_response = object_client.presign_upload(project_id, datasets_path)
 
-    for i, chunk in enumerate(_rechunk_data(content)):
+    if presign_response.provider == CloudStorageProvider.S3:
+        if known_file_size is None:
```
acroz: There's a bit of duplication between here and below (both this if statement logic plus two other places where the `DEFAULT_CHUNK_SIZE` can be returned). How about merging all these cases together in a single `_chunk_size` helper like:

```python
def _chunk_size(provider, known_file_size):
    if provider == CloudStorageProvider.S3 and known_file_size is not None:
        return max([math.ceil(total_size / S3_MAX_CHUNKS), DEFAULT_CHUNK_SIZE])
    else:
        return DEFAULT_CHUNK_SIZE
```

Then, in this function, we can just call:

```
chunk_size = _chunk_size(presign_response.provider, known_file_size)
```

in one place before the if statement.

tests/datasets/test_transfer.py
```diff
@@ -193,7 +207,7 @@ def second_chunk_request_matcher(request):
         status_code=200,
     )
 
-    mock_client_upload_s3.presign_upload_part.return_value = TEST_URL
+    # mock_client_upload_s3.presign_upload_part.return_value = TEST_URL
```
acroz: Delete this line?

tests/datasets/test_transfer.py
```diff
@@ -208,6 +222,37 @@ def second_chunk_request_matcher(request):
...
+        lambda x: next(chunks) == x.text.encode("utf-8")
+        for i in range(max_chunks)
+    ]
+    urls = [
+        "https://example.com/presigned-url-{i}/url".format(i=i)
+        for i in range(max_chunks)
+    ]
+    tags = ["tag-{tagid}".format(tagid=i) for i in range(max_chunks)]
+
+    for i, (url, cm, tag) in enumerate(zip(urls, chunk_matchers, tags)):
```
acroz: You're not using the `i` generated by `enumerate` - remove it?

Could we also rename tag -> etag and cm -> matcher (or similar), for explicitness?

tests/datasets/test_transfer.py
```diff
@@ -127,29 +134,36 @@ def test_chunking_and_labelling_data_of_exact_sizes(mocker):
...
-    chunks = transfer._rechunk_and_label_as_last(content)
+    chunks = transfer._rechunk_and_label_as_last(content, chunk_size)
     assert list(chunks) == [
         (b"1111", False),
         (b"2222", False),
         (b"last", True),
     ]
 
 
+def test_s3_dynamic_chunk(mocker):
```
acroz: It's not clear which function (`_s3_chunk_size` or `_rechunk_data`) is really being tested here. I'd prefer that only the  function under test gets called.

acroz:
This looks great - I've merged master to get the Python 3.8 checks Gergely added.

Before merging, have you validated that the new functionality works in practice against both S3 and GCS deployments, and that the change fixes the issue on S3?

jkeelan:
I'd previously checked that uploads still worked on S3, and I've now validated that a 54G file correctly uploads to S3. I don't think I have access to a GCS deployment though.

imrehg:
@jkeelan @pbugnion tested it on a GCP deployment, uploaded a 55G file totally fine.

acroz:
@imrehg Thanks for testing this out! I've rebased off master - assuming tests pass I'll merge and release.

acroz:
This is now released in 0.25.6. It'll not appear on platform server by default until we do a rebuild, but you can get the latest version by doing `pip install -U faculty`.

Issue #148, opened 2019-12-19
Can't use faculty.datasets natively with pathlib


snakch:
At the moment, it looks like we have to convert PosixPath objects to strings before passing them to datasets functions.

imrehg:
Looking into this, my impression is so far, that `pathlib.Path` is more aimed at local paths (and dependent on the local filesystem's properties, i.e. the separators, etc). Thus for remote path should likely not use that. Also does some path manipulation behind the scenes (such as trailing backslash removals) that throws a 🔧 in the machine, imho.

If you still wanna do things somewhat more programmatically, `os.path`  and friends might be good. But to be actually correct/better, use `posixpath`  (as the remote storage should be according to that, and thus no local setup can interfere). Something like:

```python
import posixpath as path
import faculty.datasets as datasets

dataset_path = path.join("/")
# This will be: /
print(datasets.ls(dataset_path))

dataset_path = path.join("/", 'something')
# This will be: /something
print(datasets.ls(dataset_path))

dataset_path = path.join("/", 'something', 'else')
# This will be: /something/else
print(datasets.ls(dataset_path))
```

Hope this helps!

PR #150, opened 2019-12-19
tests: datasets.get working with both string and posixpath

imrehg:
Connects-to: #148
Signed-off-by: Gergely Imreh <gergely.imreh@faculty.ai>

### DIFF
--- a/tests/datasets/test_init.py
+++ b/tests/datasets/test_init.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 
+import posixpath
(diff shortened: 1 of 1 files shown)


imrehg:
Actually this should be unnecessary, as the return value of `posixpath.join` is just a `str`, so no different from using a string in the first place...

```python
>>> import posixpath
>>> type(posixpath.join('/', 'test'))
<class 'str'>
``` 
Closing as not adding real value after all.

PR #152, opened 2020-01-22
Add module to make datasets IO easier with pandas

acroz:
This needs tests added before merging.

### DIFF
--- a/faculty/datasets/pandas.py
+++ b/faculty/datasets/pandas.py
@@ -1,0 +1,77 @@
+# Copyright 2018-2019 Faculty Science Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pandas
+
+from faculty.session import get_session
+from faculty.context import get_context
+from faculty.clients.object import ObjectClient
+from faculty.datasets import transfer
+
+
+def read_csv(project_path, project_id=None, *args, **kwargs):
+    """Read a CSV file from a project's datasets as a pandas DataFrame.
+
+    Requires pandas to be installed.
+
+    Parameters
+    ----------
+    project_path : str
+        The path of the CSV file in the project's datasets to read.
+    project_id : str, optional
+        The project to get the CSV file from. You need to have access to this
+        project for it to work. Defaults to the project set by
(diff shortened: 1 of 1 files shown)


imrehg:
This is quite interesting! Some initial thoughts, and coming from a place of ignorance:

* that namespacing in `faculty.datasets.pandas....` seems good, though would you think people would find it confusing that only some functions (and not all pandas) are propagated? I guess not really, but guessing it's good to also document in any case, later on.
* would we consider adding similar functions to the other `read_FORMAT` and `to_FORMAT` as well? Looking at the API reference, I can imagine `FORMAT` as in excel, json, HDF, parquet, pickle.... I bet these are less frequent, but feel like if include one, should include the rest as well.

CDonnerer:
Agreed that this seems super convenient! Comments from an even higher place of ignorance:

- Calling it `import faculty.datasets.pandas as faculty_pandas` is a bit confusing if it doesn't reproduce the whole pandas API.

- Adding to the above, how about just having generic `read` and `write` functions? Then you could interface with any type of data without adding dependencies.

```
import pandas as pd
import cv2
from faculty.datasets.io import read, write

df = read('my_path.csv', pd.read_csv)
video = read('my_path.MP4', cv2.VideoCapture)
```

imrehg:
Also just for clarity, with the 2 options above, if the 2nd is used, you mean that the whole of pandas would be available as a namespaced component? Or just these functions?

For the shorter import, I wonder if either
```
import faculty.datasets.pandas as fdpd
import faculty.datasets.pandas as fpd
```
would be more natural (so keeping the original `pd` convention, but adding some way to highlight the "faculty-ness" of things. Just a thought, no strong preference.

sbalian:
Thanks @acroz , thought about this a bit more and considered all the comments above. How about the following?

```python
from faculty import datasets

url = datasets.presigned_url("/path/to/any/file")
```

We can then add a section in the docs (or docstrings) illustrating usage with `pd.read_*`, and possibly readers from other libraries that support url inputs.

As for writing, you can do something like `datasets.put_string` to take the local file as a string (as returned by `pd.DataFrame.to_csv(path_or_buf=None)` and also `pd.Series.to_csv`), and again illustrate usage in the docs. Or perhaps modify `datasets.put` so that it can take the string as well as the file path as input. Finally, we can also have the inverse of this - `datasets.get_string`.

This satisfies the two requirements:

1. No dependence on pandas: is general and removes the burden of including pandas as a dependency.
2. Makes it easier to deal with datasets IO during development. To me, the main burden is dealing with `ObjectClient` when I want a presigned URL or when I want to upload data that is not sitting on disk.

sbalian:
@acroz Also ran a quick test to compare speed for an AWS backend.

For a 139M CSV file,

| Method | Time in seconds |
|---------|------|
|`pandas.read_csv` | 3.29 |
|`faculty.datasets.pandas.read_csv` | 5.52 |
|`pandas.DataFrame.to_csv` | 10.2|
|`faculty.datasets.pandas.to_csv` | 18.7 |

These are very promising because `object price/workspace price << 0.5`, and here we are seeing that workspace speedup over object is not even `2x`.

PR #153, opened 2020-02-03
Remove erroneous line that appears to be overridden

acroz:


acroz:
I don't believe this issue actually affects the function of the code, so I'm just going to leave to be bundled in the next release.

PR #155, opened 2020-02-18
Add ability to cancel running jobs in JobClient

imrehg:
Depends on steve version 0.1.64
Signed-off-by: Gergely Imreh <gergely.imreh@faculty.ai>

### DIFF
--- a/faculty/clients/job.py
+++ b/faculty/clients/job.py
@@ -398,6 +398,22 @@ class JobClient(BaseClient):
         )
         return self._get(endpoint, _SubrunSchema())
 
+    def cancel_run(self, project_id, job_id, run_identifier):
+        """Cancel a running job.
+
+        Parameters
+        ----------
+        project_id : uuid.UUID
+        job_id : uuid.UUID
+        run_identifier : uuid.UUID or int
+            The ID or number of the run to cancel.
+        """
+
+        endpoint = "/project/{}/job/{}/run/{}".format(
+            project_id, job_id, run_identifier
+        )
+        self._delete_raw(endpoint)
+
 
 class _JobMetadataSchema(BaseSchema):
     name = fields.String(required=True)
--- a/faculty/clients/job.py
+++ b/faculty/clients/job.py
@@ -404,7 +404,9 @@ class JobClient(BaseClient):
         Parameters
         ----------
         project_id : uuid.UUID
+            The ID of the project containing the job.
         job_id : uuid.UUID
+            The ID of the job containing the run.
         run_identifier : uuid.UUID or int
             The ID or number of the run to cancel.
         """
--- a/tests/clients/test_job.py
+++ b/tests/clients/test_job.py
@@ -709,3 +709,17 @@ def test_job_client_get_subrun(mocker, run_identifier, subrun_identifier):
         ),
         schema_mock.return_value,
     )
+
+
+@pytest.mark.parametrize(
+    "run_identifier", [RUN_ID, RUN.run_number], ids=["ID", "Number"]
+)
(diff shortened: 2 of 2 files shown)


faculty/clients/job.py
```diff
@@ -580,3 +580,19 @@ def get_subrun(
             project_id, job_id, run_identifier, subrun_identifier
         )
         return self._get(endpoint, SubrunSchema())
+
+    def cancel_run(self, project_id, job_id, run_identifier):
+        """Cancel a runnig job.
```
```suggestion
        """Cancel a running job.
```

faculty/clients/job.py
```diff
@@ -580,3 +580,19 @@ def get_subrun(
             project_id, job_id, run_identifier, subrun_identifier
         )
         return self._get(endpoint, SubrunSchema())
+
+    def cancel_run(self, project_id, job_id, run_identifier):
```
acroz: I would change `run_identifier` to `run_id`, as this endpoint just accepts the run ID (a UUID), not either the run ID or run number (an int) as other endpoints do.

tests/clients/test_job.py
```diff
@@ -709,3 +709,14 @@ def test_job_client_get_subrun(mocker, run_identifier, subrun_identifier):
         ),
         schema_mock.return_value,
     )
+
+
+def test_job_client_cancel_run(mocker, run_identifier=RUN_ID):
```
acroz: I would just embed the `RUN_ID` in the code below rather than adding this kwarg. kwargs as arguments to test functions are normally used only for providing fixtures (e.g. `mocker`) or values for parameterized tests.

imrehg: Yeah, this is a placeholder as expecting to update Steve to take both number and id, and then it will be rewritten. (draft state)

faculty/clients/job.py
imrehg: Yeah, hope to work this out to be an actual `run_identifier` indeed, caught me in the act. :)

faculty/clients/job.py
```diff
@@ -580,3 +580,19 @@ def get_subrun(
...
+        """Cancel a runnig job.
+
+        Parameters
+        ----------
+        project_id : uuid.UUID
+        job_id : uuid.UUID
+        run_identifier : uuid.UUID
```
```suggestion
        run_identifier : uuid.UUID or int
```
If this method is going to support run numbers

faculty/clients/job.py
imrehg: Depends on https://bitbucket.org/theasi/steve/pull-requests/191

acroz:
@imrehg Did the Steve version this depends on get released? If so, can we rebase this and test it against a live deployment prior to merge?

imrehg:
@acroz cheers for the bump, I've rebased it. Trying to test against Cloud, but seems like the version might not been deployed there yet. Just checking, and will follow up when I have confirmation.

faculty/clients/job.py
```diff
@@ -398,6 +398,22 @@ def get_subrun(
         )
         return self._get(endpoint, _SubrunSchema())
 
+    def cancel_run(self, project_id, job_id, run_identifier):
+        """Cancel a running job.
+
+        Parameters
+        ----------
+        project_id : uuid.UUID
```
acroz, line 406: With the merging of the docs PR, we now include descriptions for all parameters. Can you copy the project_id and job_id descriptions from other methods into this docstring?

imrehg: Sure thing, sorry, I should have looked at the other changes more closely!

imrehg:
Tested on Cloud and appears to function as expected.

acroz:
I've released this with version 0.26.1.

PR #159, opened 2020-02-28
Add documentation

acroz:
This PR adds Sphinx docs with API documentation.

As well as adding the Sphinx documentation itself, this PR adds docstrings for all public facing classes and methods, which are then used to generate API docs automatically. Non-public facing classes (schemas, a few models only used internally and helper functions) have been marked as private by the 'leading underscore' Python convention to both exclude them from the docs and better convey their status to users.

These renames will break any existing use of the affected classes and functions, however we realistically can't commit to supporting these parts that are supposed to be internals anyway.

Later PRs are planned to:

* Add more descriptive user guides which provide basic examples and help people to consume the API docs
* Change existing use of `NamedTuple`s to `@attrs` classes, partly because they're better, but also because they're easier to write documentation for.
* Add CI to build the docs and deploy them to a publicly accessible location.

I would like to merge this PR without waiting for the above pieces, as it's almost certain that other PRs on this library will conflict with this one.


pbugnion:
This is brilliant. I haven't viewed the documentation in detail, but I'm almost certain that having some documentation is better than none, so I think we should deploy it. It'll be easier to improve and tweak once it's live.

PR #161, opened 2020-03-16
Added method for listing all servers for user

jkeelan:
Added method for the "/user/<user_id>/instances" endpoint to the ServerClient.


acroz:
This looks good, but needs merge conflicts to be resolved now that the docs PR is merged.

I would also ask you to include a docstring for this method in line with the docstrings recently added to other methods.

PR #162, opened 2020-04-30
Make `datasets.mv` do nothing if source and destination are the same

liamcoatman:
Make `datasets.mv` do nothing if source and destination are the same. 

Because `mv` is implemented as `cp` then `rm`, the previous behaviour was to delete the file if the source and destination are the same. The new behaviour is to do nothing (like the Unix `mv` command).

### DIFF
--- a/faculty/datasets/__init__.py
+++ b/faculty/datasets/__init__.py
@@ -334,6 +334,9 @@ def mv(source_path, destination_path, project_id=None, object_client=None):
     project_id = project_id or get_context().project_id
     object_client = object_client or ObjectClient(get_session())
 
+    if source_path == destination_path:
+        return
+
     cp(
         source_path,
         destination_path,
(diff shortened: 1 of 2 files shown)


acroz:
I suggest we close this in favour of #163. Then, it's the backend's responsibility to get this behaviour right. If the backend implements this behaviour wrongly, we should fix it. Thoughts?

liamcoatman:
> I suggest we close this in favour of #163. Then, it's the backend's responsibility to get this behaviour right. If the backend implements this behaviour wrongly, we should fix it. Thoughts?

Agreed.

jkeelan:
I disagree, I think we should push this, because it prevents an unexpected and potentially destructive operation. To be clear, #163 also suffers from this issue, so it would have to be fixed on the backend, which will mean a further delay. While solving it on the front end isn't ideal, it's better than nothing.

acroz:
Fair enough - let's merge this and deal with the backend implementation separately.

acroz:
Released with [0.26.2](https://pypi.org/project/faculty/0.26.2/)

PR #163, opened 2020-04-30
Add move to object client + change datasets.mv

jkeelan:
This PR adds move to the object client, using the `object-move` endpoint.

This reduces the number of API calls for `datasets.mv`. I've tested the functionality and it seems to work as expected.


imrehg:
Can this and #162 combined maybe? Or either way have to keep that other one in mind for this change too, right?

jkeelan:
Yeah. Either combine or I'll merge once #162 hits master.

PR #164, opened 2020-04-30
Add stat function to datasets

Matt-Haugh:
This PR adds a stat function to datasets which mimics some of the functionality of https://docs.python.org/3/library/os.html#os.stat.


Matt-Haugh:
Cheers @acroz, that sounds sensible. In fact part of the motivation for adding this was to add an `ls -l` to the datasets branch that @liamcoatman has for faculty-cli. I'll get on reworking this over the next few days

PR #166, opened 2020-05-06
Test updates

acroz:
* Switch CI to GitHub
* Remove unused setuptools alias
* Don't test on Python 3.4, which is no longer supported
* Don't check flake8 on Python 2 - Python 2 support will be removed soon and Python 2-specific flake8 failures should not prevent merging


acroz:
The repo is still trying to do the checks with Travis, but I hope that after merging this PR, I can disable the Travis integration and that will go away.

PR #167, opened 2020-05-07
Merge GitHub workflows into single file

acroz:
This will allow replacing the Travis badge with a single GitHub one,
and I think this is also closer to the intended way to organise things.


imrehg:
Looking good!

Screenshot 2020-05-07 15 49 00

PR #170, opened 2020-09-04
Publish new template

tomas-milata:
This PR adds 2 clients that we need to implement the `faculty template new` CLI command (see the [CLI changes](https://github.com/facultyai/faculty-cli/pull/43)).

The template client that talks to `kanto` is quite straightforward.

The notification client is a bit different from clients for other services because it uses server-sent events

I've used an external library for SSEs. We have an existing code snippet for an SSE client in `faculty-cli` I could copy, but it didn't seem it was worth debugging much. The library works and by looking at the source code it seems more robust than our minimalistic implementation.

This depends on exposing frontend on `https://frontend.services.???.my.faculty.ai` and also allowing SSEs from `frontend.sml-services` (for which I have changes ready and validated it works):
Screenshot 2020-09-10 at 11 00 16

### DIFF
--- a/faculty/clients/frontend.py
+++ b/faculty/clients/frontend.py
@@ -30,20 +30,6 @@ class FrontendClient(BaseClient):
 
     _SERVICE_NAME = "frontend"
 
-    # TODO this is a copy-paste modificiation from BaseClient
-    def _request(self, method, endpoint, check_status=True, *args, **kwargs):
-        """Perform an HTTP request.
-
-        This method should not be called from subclasses directly. Instead,
-        call one of the HTTP verb-specific methods. If it does not exist yet
-        for the HTTP method you need, contribute it.
-        """
-        url = "https://gollum.platform.asidata.science/" + endpoint 
-        response = self.http_session.request(method, url, *args, **kwargs)
-        if check_status:
-            _check_status(response)
-        return response
-
     def user_updates(self, user_id):
         endpoint = "api/updates/user/{}".format(user_id)
         response = self._get_raw(endpoint, stream=True)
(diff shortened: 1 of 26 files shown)


faculty/clients/frontend.py
```diff
@@ -0,0 +1,65 @@
...
+import json
+
+import requests
+import sseclient
+
+from faculty.clients.auth import FacultyAuth
+from faculty.clients.base import _check_status, BaseClient
+import faculty.session
+
+class FrontendClient(BaseClient):
```
acroz, line 29: Given that we're unlikely to use this client for anything else, and we can have multiple clients with the same `_SERVICE_NAME` field if we really need to, would it make sense to have the naming of this be something like `NotificationClient` or `PlatformEventClient` etc.?

tomas-milata:
Hang on with the review, I think we'll change the client to use `frontend.services.XX.my.faculty.ai` after all.

tomas-milata:
> Hang on with the review, I think we'll change the client to use `frontend.services.XX.my.faculty.ai` after all.

This is now ready to review.

faculty/clients/notification.py
```diff
@@ -0,0 +1,102 @@
...
+        ----------
+        project_id : uuid.UUID
+            The project from which the template was published.
+        events : generator
+            The value that was returned from
+            :func:`faculty.clients.notification.NotificationClient.user_updates`
+        """
+        for e in events:
+            body = json.loads(e.data)
+            if body["sourceProjectId"] == str(project_id):
```
zblz, line 76: We probably shouldn't fail here if an event does not have a `sourceProjectId` field:
```suggestion
            if body.get("sourceProjectId") == str(project_id):
```

PR #172, opened 2020-09-10
Publish new template

tomas-milata:
(Reopening https://github.com/facultyai/faculty/pull/170 which I accidentally merged 🤦 )

This PR adds 2 clients that we need to implement the faculty template new CLI command (see the CLI changes).

The template client that talks to kanto is quite straightforward.

The notification client is a bit different from clients for other services because it uses server-sent events

I've used an external library for SSEs. We have an existing code snippet for an SSE client in faculty-cli I could copy, but it didn't seem it was worth debugging much. The library works and by looking at the source code it seems more robust than our minimalistic implementation.

This depends on exposing frontend on https://frontend.services.???.my.faculty.ai and also allowing SSEs from frontend.sml-services (for which I have changes ready and validated it works):
Screenshot 2020-09-10 at 11 00 16

### DIFF
--- a/faculty/clients/notification.py
+++ b/faculty/clients/notification.py
@@ -1,0 +1,102 @@
+# Copyright 2020 Faculty Science Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Consume notifications from the Faculty frontend.
+"""
+
+import json
+
+import sseclient
+
+from faculty.clients.base import BaseClient
+
+
+class NotificationClient(BaseClient):
+    """Client to listen on notifications from the Faculty frontend.
+
+    Either build this client with a session directly, or use the
+    :func:`faculty.client` helper function:
+
+    >>> client = faculty.client("notification")
+
+    Parameters
+    ----------
+    session : faculty.session.Session
+        The session to use to make requests
+    """
+
+    _SERVICE_NAME = "frontend"
+
+    def user_updates(self, user_id):
+        """Get notification events for the given user.
+
+        Parameters
+        ----------
+        user_id : uuid.UUID
+            ID of the user to get updates for.
+
+        Returns
+        -------
+        generator
+            Server-sent events
+        """
+        endpoint = "api/updates/user/{}".format(user_id)
+        response = self._get_raw(endpoint, stream=True)
+        client = sseclient.SSEClient(response)
+        return client.events()
+
+    def check_publish_template_result(self, events, project_id):
+        """Handle results of a template publishing operation.
+
+        Only returns when success or failure events are received for the given
+        source project_id. Events with other project IDs are ignored.
+
+        Parameters
+        ----------
+        project_id : uuid.UUID
+            The project from which the template was published.
+        events : generator
+            The value that was returned from
+            :func:`faculty.clients.notification.NotificationClient.user_updates`
+        """
+        for e in events:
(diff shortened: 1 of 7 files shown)


faculty/clients/template.py
```diff
@@ -0,0 +1,45 @@
...
+
+    Parameters
+    ----------
+    session : faculty.session.Session
+        The session to use to make requests
+    """
+
+    _SERVICE_NAME = "kanto"
+
+    def publish_new(self, template, source_directory, project_id):
```
acroz, line 38: Nit: We usually put the project_id first in client methods.

faculty/clients/notification.py
```diff
@@ -0,0 +1,102 @@
...
+        -------
+        generator
+            Server-sent events
+        """
+        endpoint = "api/updates/user/{}".format(user_id)
+        response = self._get_raw(endpoint, stream=True)
+        client = sseclient.SSEClient(response)
+        return client.events()
+
+    def check_publish_template_result(self, events, project_id):
```
acroz, line 60: Are we assuming that only one template is published at a time on a project? What if two users publish templates from the same project at similar times?

acroz, line 60: It does feel a little odd that what's essentially a helper function is exposed as a method on this class. Would it make more sense for it to call `self.user_updates()` internally and filter the returned event stream here? If you want to facilitate testing, it's possible to mock `user_updates` on a real instance of `NotificationClient`.

acroz:
Looks good, left a few comments.

Is this supposed to target `master`?

faculty/clients/notification.py
tomas-milata: We assume 1 template is published at the same time for 1 (project, user).

The `event` passed to `check_publish_template_result` are meant to be those returned by `user_updates` which are for 1 user.

Yes, this could break if 1 user published 2 templates at once. I think frontend would break in the same way, it also just listens to the (project+user) updates. I considered handling this by generating a unique request ID for later checks, but then decided it wasn't worth the hassle given how likely I think this is going to be used and how likely it is for 1 user to run publishing more than once at a time.

tomas-milata: The reason for the split is to be able to do this sequence of things:

1) start listening on events (`user_updates` returns a generator).
2) run publishing in kanto
3) handle the incoming events

If 2) was run before 1), we'd miss events. (I've tried that).

(Not sure if there's an alternative way to implement this instead of a generator, but I wanted to avoid some asyncio hackery, given we haven't dropped python2 yet. And this seems to work).

tomas-milata:
> Looks good, left a few comments.
> 
> Is this supposed to target `master`?

Probably not, I've changed it to a dev branch which we'll keep until all publishing is done. Thanks.

faculty/clients/notification.py
acroz: Hmm, ok.

acroz: Ah, I missed the fact that the method you want to block on is itself not a generator.

You could wrap the event stream (generator) in a class that implements the "block until complete", e.g.:

```
class PublishTemplateNotifications:
    def __init__(self, events):
        self.events = events

    def wait_for_completion(self):
        for e in self.events:
            if e.event == "@SSE/PROJECT_TEMPLATE_PUBLISH_NEW_FAILED":
                body = json.loads(e.data)
                msg = _extract_publishing_error_msg(body)
                raise TemplatePublishingError(msg)
            elif e.event == "@SSE/PROJECT_TEMPLATE_PUBLISH_NEW_COMPLETED":
                return


class NotificationClient:


    def user_updates(self, user_id):
        ...

    def publish_template_notifications(self, user_id, source_project_id):

        def condition(event):
            if not event.event.startswith("@SSE/PROJECT_TEMPLATE_PUBLISH"):
                return False
            body = json.loads(event.data)
            return body.get("sourceProjectId") == str(source_project_id)

        return PublishTemplateNotifications(
            filter(condition, self.user_updates(user_id))
        )
```

tomas-milata: Hmm, I think I find your solution more difficult to understand because the publishing-specific logic is split into 2 places and I'm also not very used to putting logic like `wait_for_completion` into something that seems basically like a model class (`PublishTemplateNotification`). But it's fair to say that in your solution it is clearer where the blocking happens.

So I think I'll trust your judgement.

PR #175, opened 2020-09-21
Add template to project from directory

tomas-milata:
This adds clients that are needed for the `faculty template add-to-project-from-directory` CLI command (see the [PR](https://github.com/facultyai/faculty-cli/pull/44).

It also unifies some of the error handling logic because the backend responses are almost identical as in the `template publish new` flow (except for some HTTP 409 vs 400 differences, only relevant for frontend).

There is missing error handling logic for the error notifications -- needs to be added to backend yet.

### DIFF
--- a/faculty/clients/notification.py
+++ b/faculty/clients/notification.py
@@ -178,7 +178,7 @@ class AddTemplateToProjectFromDirectoryNotifications:
                 e.event == "@SSE/PROJECT_TEMPLATE_APPLY_FROM_DIRECTORY"
                 "_ADD_TO_PROJECT_FAILED"
             ):
-                body = json.loads(e.data)
+                json.loads(e.data)
                 # TODO backend needs to return rendering errors here
                 msg = "Failed to apply template to directory"
                 raise TemplatePublishingError(msg)
--- a/faculty/clients/notification.py
+++ b/faculty/clients/notification.py
@@ -179,7 +179,8 @@ class AddTemplateToProjectFromDirectoryNotifications:
                 "_ADD_TO_PROJECT_FAILED"
             ):
                 body = json.loads(e.data)
-                msg = _extract_publishing_error_msg(body)
+                # TODO backend needs to return rendering errors here
+                msg = "Failed to apply template to directory"
                 raise TemplatePublishingError(msg)
             elif (
                 e.event == "@SSE/PROJECT_TEMPLATE_APPLY_FROM_DIRECTORY"
--- a/faculty/clients/notification.py
+++ b/faculty/clients/notification.py
@@ -57,7 +57,7 @@ class NotificationClient(BaseClient):
         client = sseclient.SSEClient(response)
         return client.events()
 
-    def get_publish_template_notifications(self, user_id, source_project_id):
+    def publish_template_notifications(self, user_id, source_project_id):
         """Get notification events of a template publishing operation.
 
         Only returns when success or failure events are received for the given
@@ -78,6 +78,34 @@ class NotificationClient(BaseClient):
             return body.get("sourceProjectId") == str(source_project_id)
 
         return PublishTemplateNotifications(
+            filter(is_publishing_event, self.user_updates(user_id))
+        )
+
+    def add_to_project_from_dir_notifications(
+        self, user_id, target_project_id
+    ):
+        """Get notifications of a "add to project from directory" operation.
+
+        Only returns when success or failure events are received for the given
+        target project_id. Events with other project IDs are ignored.
+
+        Parameters
+        ----------
+        user_id : uuid.UUID
+            The user who started the publish operation.
+        target_project_id : uuid.UUID
+            The project from which the template was published.
+        """
+
+        def is_publishing_event(event):
+            if not event.event.startswith(
+                "@SSE/PROJECT_TEMPLATE_APPLY_FROM_DIRECTORY_ADD_TO_PROJECT"
+            ):
+                return False
+            body = json.loads(event.data)
+            return body.get("projectId") == str(target_project_id)
+
+        return AddTemplateToProjectFromDirectoryNotifications(
             filter(is_publishing_event, self.user_updates(user_id))
         )
 
@@ -105,7 +133,7 @@ class TemplatePublishingError(Exception):
 
 class PublishTemplateNotifications:
     """
-    Wrapper around notification events from template publishig.
+    Wrapper around notification events from template publishing.
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


faculty/clients/template.py
```diff
@@ -42,4 +45,310 @@ def publish_new(self, project_id, template, source_directory):
...
+            raise ParameterValidationFailureSchema().load(response_body)
+        elif error_code == "template_retrieval_failure":
+            raise TemplateRetrievalFailureResponseSchema().load(response_body)
+        elif error_code == "default_parameters_parsing_error":
+            raise DefaultParametersParsingErrorSchema().load(response_body)
+        elif error_code == "generic_parsing_failure":
+            raise GenericParsingErrorSchema().load(response_body)
+        elif error_code == "workspace_files_validation_error":
+            raise WorkspaceFilesValidationErrorSchema().load(response_body)
+    else:
```
mociarain, line 88: There's the ServiceUnavailable response too, which this would swallow. Not sure we care but it's not an "Unexpected response".

tomas-milata: Yeah, I wasn't sure it's worth keeping this response in the backend and was planning to remove it (it's actually not present in the `publish new` flow. I'm not convinced it's worth the hassle to propagate this error if the server cant' process the request (compared to just throwing and IllegalStateException in the server and returning a 500).

mociarain (approved):
Just left 1 question but LGTM

faculty/clients/notification.py
```diff
@@ -126,3 +154,36 @@ def wait_for_completion(self):
...
+
+class AddTemplateToProjectFromDirectoryNotifications:
+    """
+    Wrapper around notification events from the "add template to project from
+    directory" operation
+
+    Parameters
+    ----------
+    events : generator
+            Events from the "add to project from directory" operation.
```
acroz, line 167: "Events" should only be indented by 4 spaces compared to the previous line.

faculty/clients/notification.py
```diff
@@ -126,3 +154,36 @@ def wait_for_completion(self):
                 raise TemplatePublishingError(msg)
             elif e.event == "@SSE/PROJECT_TEMPLATE_PUBLISH_NEW_COMPLETED":
                 return
+
+
+class AddTemplateToProjectFromDirectoryNotifications:
+    """
+    Wrapper around notification events from the "add template to project from
```
acroz, line 161: Python docstrings should start with a single-line description, with further elaboration, if required, put separately in paragraphs, e.g.:

```
"""
A function to do a thing.

Lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem ipsum
lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem ipsum
lorem ipsum lorem ipsum lorem ipsum.

Lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem ipsum
lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem ipsum
lorem ipsum lorem ipsum lorem ipsum.

Parameters
----------
...
```

faculty/clients/template.py
```diff
@@ -42,4 +45,310 @@ def publish_new(self, project_id, template, source_directory):
...
+        raise Exception(
```
acroz, line 89: Rather than raising a vanilla `Exception`, it would be better if we could surface the custom exceptions that the library would normally raise. This could be done with a small tweak to your logic - instead of passing `check_status=False`, allow `_post_raw` to raise an exception, handle recognised cases, and re-raise any unhandled cases:

```
from faculty.clients.base import HttpError

...

    def add_to_project_from_directory(...):
        endpoint = "project/{}/apply".format(target_project_id)
        payload = {...}
        try:
            self._post_raw(endpoint, json=payload)
        except HttpError as err:
            _handle_publishing_error(err)

def _handle_publishing_error(exception):
    if exception.error_code == "resources_validation_failure":
        raise ResourceValidationFailureResponseSchema().load(exception.response.json())
    elif exception.error_code == "parameter_validation_failure":
        raise ParameterValidationFailureSchema().load(exception.response.json())
    elif exception.error_code == "template_retrieval_failure":
        raise TemplateRetrievalFailureResponseSchema().load(exception.response.json())
    elif exception.error_code == "default_parameters_parsing_error":
        raise DefaultParametersParsingErrorSchema().load(exception.response.json())
    elif exception.error_code == "generic_parsing_failure":
        raise GenericParsingErrorSchema().load(exception.response.json())
    elif exception.error_code == "workspace_files_validation_error":
        raise WorkspaceFilesValidationErrorSchema().load(exception.response.json())
    else:
        raise
```

faculty/clients/template.py
```diff
@@ -42,4 +45,310 @@ def publish_new(self, project_id, template, source_directory):
...
+            "Unexpected response from the server:\n", response.text
+        )
+
+
+class TemplateException(Exception):
+    pass
+
+
+FileTooLargeError = namedtuple(
```
acroz, line 98: In new clients, we're preferring to use `attrs` instead of named tuples for data models. This is probably not worth going back and changing in this PR, but mentioning it for future reference. See `faculty/clients/model.py` for examples. Also FAO @mociarain.

acroz: Actually, in this particular case this shouldn't be a namedtuple as it seems to be intended to be an exception. In this case, `FileTooLargeError` should inherit from `Exception` to behave properly as an exception.

FWIW, this is easy to achieve with attrs:

```
from attr import attrs, attrib

@attrs
class FileTooLargeError(Exception):
    path = attrib()
    actual_size_bytes = attrib()
    max_bytes = attrib()
```

acroz: This also applies to `TooManyFilesError` below.

acroz: Oh, never mind - I see that these are actually nested in the `WorkpaceFilesValidationError` class. Perhaps these named tuples should not have `Error` in the name to avoid suggesting they are themselves exceptions?

faculty/clients/template.py
```diff
@@ -42,4 +45,310 @@ def publish_new(self, project_id, template, source_directory):
...
+
+class WorkspaceFilesValidationErrorSchema(BaseSchema):
+    files_too_large = fields.List(
+        fields.Nested(FileTooLargeSchema()),
+        data_key="filesTooLarge",
+        required=True,
+    )
+    too_many_files = fields.Nested(
+        TooManyFilesSchema(), data_key="tooManyFiles"
+    )
```
acroz, line 141: Is it correct that `filesTooLarge` is always set and that `tooManyFiles` is always optional? I don't know enough about the data model to know, but the discrepancy seems odd.

faculty/clients/template.py
```diff
@@ -42,4 +45,310 @@ def publish_new(self, project_id, template, source_directory):
...
+
+    @staticmethod
+    def from_errors(errors):
+        return ResourceValidationFailure(
+            apps=errors["apps"],
+            apis=errors["apis"],
+            environments=errors["environments"],
+            jobs=errors["jobs"],
+            workspace=errors["workspace"],
+        )
```
acroz, line 309: Weak suggestion avoiding the need for the `classmethod`:

```suggestion
class ResourceValidationFailureResponseSchema(BaseSchema):
    errors = fields.Nested(ResourceValidationFailuresSchema(), required=True)

    @post_load
    def make_failure(self, data, **kwargs):
        errors = data["errors"]
        return ResourceValidationFailure.from_errors(**errors)


class ResourceValidationFailure(TemplateException):
    def __init__(self, apps, apis, environments, jobs, workspace):
        self.apps = apps
        self.apis = apis
        self.environments = environments
        self.jobs = jobs
        self.workspace = workspace
```

faculty/clients/template.py
```diff
@@ -42,4 +45,310 @@ def publish_new(self, project_id, template, source_directory):
...
+    environments = fields.List(fields.String(), required=True)
+    jobs = fields.List(fields.String(), required=True)
+
+
+class TemplateRetrievalFailureResponseSchema(BaseSchema):
+    errors = fields.Nested(TemplateRetrievalFailureSchema(), required=True)
+
+    @post_load()
+    def make_failure(self, data, **kwargs):
+        return TemplateRetrievalFailure.from_errors(**data)
```
acroz, line 354: Above suggestion WRT removing classmethod applies here too.

acroz:
LGTM, left a few comments - nothing major.

faculty/clients/template.py
tomas-milata: Yeah, renaming is a good call.

faculty/clients/template.py
tomas-milata: That's correct. `tooManyFiles` is optional. `filesTooLarge` is a list, so the key is always present. No large files are represented as an empty list.

faculty/clients/template.py
tomas-milata: I think you meant to call the constructor on line 313? Not `from_errors`(because you just removed it). 

If so, then yes, that makes sense, thanks.

faculty/clients/template.py
tomas-milata: I thought this wasn't possible because I didn't realise the HTTP errors from the BaseClient still contain the original `response`. Good call, thanks.

PR #176, opened 2020-10-06
Add JobClient.update_definition

zblz:


mociarain (approved):
This LGTM but worth mentioning that I noticed that in the other examples of create/update we use:
- in environment.py we use `_EnvironmentCreateUpdateSchema`.
- in experiment.py we just construct the payloads manually.

Might be worth changing `_JobDefinitionSchema` -> `_JobDefinitionCreateUpdateSchema` to be consistent but it's a bit long.

acroz (approved):

acroz:
@zblz Is it useful to release this straight away? Happy to do so but if you're not using this directly (is this one of the methods you're vendoring in the airflow plugin?) I might not release so as not to confuse with the dev tags Tomas has been pushing.

PR #178, opened 2020-11-23
Unpin marshmallow version for Python 3

### LINKED ISSUE #177: Pin on marshmallow version prevents from using other packages in environment
We are currently pinning marshmallow to 3.0.0rc3, which IIRC is the last version with Python2 support. This prevents a user from using many of the marshmallow ecosystem packages that have a dependency `>=3.0.0` on marshmallow at the same time as having the faculty library installed (e.g., [`marshmallow-dataclass`](https://github.com/lovasoa/marshmallow_dataclass) or our own [`marshmallow-dataframe`](https://github.com/facultyai/marshmallow-dataframe)). At best, this will be flagged by proper dependency resolution tools like poetry or pip-compile, at worst it will cause a runtime error if a later version of marshmallow is installed.

If we don't want to drop Python 2.7 yet, we could set up the dependency like this:
```
 "marshmallow[reco]==3.0.0rc3; python_version<'3.5'",
 "marshmallow; python_version>='3.5'",
```
and update the Schemas for later versions of marshmallow (I think it passes extra kwargs to post_load) in an as backwards compatible as possible way.

zblz:
This PR keeps marshmallow pinned for Python 2 but unpins it for Python 3.

Tests have been adapted and some (which test functionality removed from mm) skipped in Python 3.

Fixes #177

### DIFF
--- a/faculty/_oneofschema.py
+++ b/faculty/_oneofschema.py
@@ -44,12 +44,12 @@ class OneOfSchema(Schema):
         class FooSchema(marshmallow.Schema):
             foo = marshmallow.fields.String(required=True)
             @marshmallow.post_load
-            def make_foo(self, data):
+            def make_foo(self, data, **kwargs):
                 return Foo(**data)
         class BarSchema(marshmallow.Schema):
             bar = marshmallow.fields.Integer(required=True)
             @marshmallow.post_load
-            def make_bar(self, data):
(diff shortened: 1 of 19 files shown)


acroz (approved):

acroz:
Released with [0.27.0](https://pypi.org/project/faculty/0.27.0/).

PR #180, opened 2020-12-17
Add methods to the account, project and server clients

acroz:
These additions facilitate updating some internal tooling, in particular removing a dependency on a legacy codebase.

### DIFF
--- a/faculty/clients/project.py
+++ b/faculty/clients/project.py
@@ -130,6 +130,25 @@ class ProjectClient(BaseClient):
         endpoint = "/user/{}".format(user_id)
         return self._get(endpoint, _ProjectSchema(many=True))
 
+    def list_all(self, include_archived=False):
+        """List all projects on the Faculty deployment.
+
+        This method requires administrative privileges not available to most
+        users.
+
+        Parameters
+        ----------
+        include_archived : bool, optional
+            If True, return archived projects. Default: False.
+
+        Returns
+        -------
+        List[Project]
+            The projects in Faculty.
+        """
+        params = {"includeArchived": include_archived}
+        return self._get("/project", _ProjectSchema(many=True), params=params)
+
 
 class _ProjectSchema(BaseSchema):
 
--- a/faculty/clients/project.py
+++ b/faculty/clients/project.py
@@ -146,7 +146,7 @@ class ProjectClient(BaseClient):
         List[Project]
             The projects in Faculty.
         """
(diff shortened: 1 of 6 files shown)


imrehg (approved):
LGTM, though not sure I'm qualified to comment on the correctness of the API calls themselves (the fields that are returned)

One question, the new `list_all` call has this note:

> This method requires administrative privileges not available to most users.

Would it be needed & possible to test the call with an unprivileged user as well? Or since the tests here are mocked, testing that is out of scope?

acroz:
@imrehg I've tested the integration against the actual backend services myself. You're right though that the automated tests in this repo mock out the actual call to the backend and so nothing special is required in this case.

faculty/clients/project.py
```diff
@@ -16,15 +16,34 @@
...
+
+    Parameters
+    ----------
+    id : uuid.UUID
+        The ID of the project.
+    name : str
+        The name of the project.
+    owner_id : uuid.UUID
+        The user ID of the owner of the project.
+    archived_at : datetime, optional
```
tomas-milata, line 38: Should this be `datetime.datetime` given that we use the full name for e.g. `uuid.UUID`?

faculty/clients/project.py
```diff
@@ -115,12 +134,32 @@ def list_accessible_by_user(self, user_id):
...
+        include_archived : bool, optional
+            If True, return archived projects. Default: False.
+
+        Returns
+        -------
+        List[Project]
+            The projects in Faculty.
+        """
+        params = {"includeArchived": int(include_archived)}
+        return self._get("/project", _ProjectSchema(many=True), params=params)
```
tomas-milata, line 154: AFAIK, the backend accepts something like `GET /project?includeArchived=true`. Is this the resulting URL? Is there something about some bool/int conversion I'm missing?

tomas-milata (commented):
Had some questions but LGTM otherwise!

faculty/clients/project.py
acroz: The backend accepts "true", "false", "1" or "0", so casting the bools to ints seemed most pragmatic.

tomas-milata:
Oh, I didn't know Play worked that way. 👍
<https://emojipedia.org/thumbs-up/>

faculty/clients/project.py
tomas-milata: Oh, I didn't know Play worked that way. 👍

PR #187, opened 2021-03-02
Pass in URL to clients instead of calculating it inside

acroz:
Currently, all clients call the session with their corresponding service name to work out the right URL. However, for the Hound service, which has a different URL for each server, does not fit in this model.

This PR brings the responsibility for determining the service URL out of the client and into the `faculty.client` helper. Most uses of the client (constructed with the helper) will remain unaffected, however direct construction of clients will need to be updated to pass the service URL.

This update will mean that an upcoming Hound client will fit in the existing pattern, avoiding the need for either two types of client or some internal complexity managing the two cases.


mociarain (approved):
LGTM. I thought it might be worth testing the `endpoint_url` composition in `base.py` but it's probably fine.

tests/clients/test_base.py
```diff
@@ -35,9 +35,9 @@
     _ErrorSchema,
 )
 
-MOCK_SERVICE_NAME = "test-service"
+MOCK_SERVICE_URL = "https://test-service.example.com/"
 MOCK_ENDPOINT = "/endpoint"
-MOCK_SERVICE_URL = "https://test-service.example.com/endpoint"
+MOCK_ENDPOINT_URL = "https://test-service.example.com/endpoint"
```
imrehg, line 40: **Nit**: Since this value derived from the previous 2 lines, would it make sense to actually derive it here along the lines how the url is calculated inside the BaseClient? I am a bit torn between readability of code (as it is), and clarity of intent (defining this line in terms of the previous two), something like:

```suggestion
MOCK_ENDPOINT_URL = MOCK_SERVICE_URL.rstrip("/") + "/" + MOCK_ENDPOINT.lstrip("/")
```

imrehg (approved):
LGTM, makes sense with the rationale as well.

tests/clients/test_base.py
acroz: Good suggestion, though I think in this case I think I prefer the fixed strings. My reasoning is that your suggestion would re-implement the logic that's under test, which I think is best avoided when possible. In this case, we validate that a particular input results in a particular output, which feels more in the spirit of unit tests.

PR #189, opened 2021-05-12
removed support for Python 2.7 and 3.5

taraganepola:
Installation now requires Python 3.6 or above.


setup.py
```diff
@@ -35,6 +35,7 @@ def load_readme():
     packages=find_packages(),
     use_scm_version={"version_scheme": "post-release"},
     setup_requires=["setuptools_scm"],
+    python_requires=">3.6",
```
acroz, line 38:
```suggestion
    python_requires=">=3.6",
```

acroz (approved):

PR #190, opened 2021-05-14
Remove Python 2 compatability

acroz:
#189 Dropped support for Python 2 - this PR removes some now-redundant code, dependencies and configuration that existed to support Python 2.


zblz (approved):
LGTM, so glad to get rid of the ridiculous marshmallow pinning!

PR #193, opened 2021-06-23
Fix datasets

acroz:
Recent changes to the client APIs broke the datasets module.

This change fixes that.


imrehg (approved):
LGTM

mociarain (commented):
LGTM

tomas-milata (approved):

PR #194, opened 2021-09-28
job definitions no longer return Conda environments, so do not require them

imrehg:
This seems to be a divergence between server responses and this library. An example server response for a Python job is as follows:

```
{
    "jobId": "[snip]",
    "meta": {
        "name": "somejob",
        "description": "",
        "authorId": "[snip]",
        "createdAt": "2021-09-28T11:10:36.321Z",
        "lastUpdatedAt": "2021-09-28T11:10:36.321Z",
    },
    "definition": {
        "workingDir": "/project",
        "command": {"name": "", "parameters": []},
        "imageType": "python",
        "instanceSizeType": "custom",
        "instanceSize": {"milliCpus": 1000, "memoryMb": 4000},
        "maxRuntimeSeconds": 3600,
        "environmentIds": [],
    },
    "createdFromProjectTemplate": None,
}
```

PR #195, opened 2021-09-28
Use attrs instead of namedtuple in jobs

acroz:
#194 made a breaking change to one of the models in the jobs client, plus I noticed that some of the models had linting errors (from mypy I think), so it seemed like a good time to bring this more up to date.


imrehg (approved):
Looks quite a lot neater!

PR #197, opened 2021-12-22
clients: add API client to this SDK

imrehg:
**This PR is a WIP**, opening it for early feedback, will update description when closer to completion.

### DIFF
--- a/faculty/clients/api.py
+++ b/faculty/clients/api.py
@@ -1,0 +1,628 @@
+# Copyright 2018-2021 Faculty Science Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Manage Faculty APIs.
+"""
+from collections import namedtuple
+from enum import Enum
+
+from marshmallow import fields, post_load
+from marshmallow_enum import EnumField
+
+from faculty.clients.base import BaseClient, BaseSchema
+from faculty.clients.server import _ServerSchema as ServerSchema
+
+Instance = namedtuple("Instance", ["instance", "outdated", "key"])
+InstanceSize = namedtuple("InstanceSize", ["milli_cpus", "memory_mb"])
+DevInstance = namedtuple("DevInstance", ["instance", "key", "instance_id"])
+DevInstanceId = namedtuple("DevInstanceId", ["instance_id"])
+ResourceLimit = namedtuple(
+    "ResourceLimit",
+    [
+        "allowed_max_milli_cpus",
+        "allowed_max_memory_mb",
+        "remaining_milli_cpus",
+        "remaining_memory_mb",
+    ],
+)
+APIInstance = namedtuple(
+    "APIInstance",
+    [
+        "project_id",
+        "api_id",
+        "error_code",
+        "error",
+        "remaining_resource_limits",
+    ],
+)
+
+APIKey = namedtuple("APIKey", ["id", "material", "enabled", "label"])
+ServerType = namedtuple("ServerType", ["instance_size_type", "instance_size"])
+WSGIDefinition = namedtuple(
+    "CommandDefinition",
+    [
+        "api_type",
+        "working_directory",
+        "conda_environment",
+        "module",
+        "wsgi_object",
+    ],
+)
+PlumberDefinition = namedtuple(
+    "PlumberDefinition", ["api_type", "working_director", "script_name"]
+)
+ScriptDefinition = namedtuple(
+    "PlumberDefinition", ["api_type", "working_director", "script_name"]
+)
+
+APIUpdate = namedtuple(
+    "APIUpdate", ["definition", "environment_ids", "default_server_size"]
+)
+
+APIDefinition = namedtuple(
+    "APIDefinition", ["name", "subdomain", "description", "last_updated_at"]
+)
+
+API = namedtuple(
+    "API",
+    [
+        "author_id",
+        "created_at",
+        "default_server_size",
+        "definition",
+        "deployment_status",
+        "description",
+        "dev_instances",
+        "environment_ids",
+        "id",
+        "last_deployed_at",
+        "last_deployed_by",
+        "name",
+        "prod_instances",
+        "prod_keys",
+        "subdomain",
+    ],
+)
+
+
+class _APIKeySchema(BaseSchema):
+    id = fields.UUID(data_key="keyId", required=True)
+    material = fields.String(required=True)
+    # Development instance keys do not have these fields
+    enabled = fields.Boolean(load_default=None)
+    label = fields.String(load_default=None)
+
+    @post_load
+    def make_apikey(self, data, **kwargs):
+        return APIKey(**data)
+
+
+class RemainingResourceLimitSchema(BaseSchema):
+    allowed_max_milli_cpus = fields.Integer(
+        data_key="allowedMaxMilliCpus", required=True
+    )
+    allowed_max_memory_mb = fields.Integer(
+        data_key="allowedMaxMemoryMb", required=True
+    )
+    remaining_milli_cpus = fields.Integer(
+        data_key="remainingMilliCpus", required=True
+    )
+    remaining_memory_mb = fields.Integer(
+        data_key="remaininigMemoryMb", required=True
+    )
+
+
+class DevInstanceIdSchema(BaseSchema):
+    instance_id = fields.UUID(data_key="instanceId")
+
+
+class APIInstanceResponseSchema(BaseSchema):
+    project_id = fields.UUID(data_key="projectId")
+    api_id = fields.UUID(data_key="apiId")
+    error_code = fields.String(data_key="errorCode")
+    error = fields.String()
+    remaining_resource_limits = fields.Nested(
+        RemainingResourceLimitSchema, data_key="remainingResourceLimits"
+    )
+
+
+class APIDevInstanceResponseSchema(APIInstanceResponseSchema):
+    instance = fields.Nested(DevInstanceIdSchema)
+    key = fields.Nested(_APIKeySchema)
+    instance_id = fields.UUID(data_key="instanceId")
+
+
+class InstanceSizeSchema(BaseSchema):
+    milli_cpus = fields.Integer(data_key="milliCpus", required=True)
+    memory_mb = fields.Integer(data_key="memoryMb", required=True)
+
+
+class APIType(Enum):
+    WSGI = "wsgi"
+    SCRIPT = "script"
+    PLUMBER = "plumber"
+
+
+class DeploymentStatus(Enum):
+    NOTDEPLOYED = "not-deployed"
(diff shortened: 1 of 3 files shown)


faculty/clients/api.py
```diff
@@ -0,0 +1,635 @@
...
+
+        payload["environmentIds"] = environment_ids
+
+        # TODO: turn these into schema handling as well
+        if default_server_size is None:
+            default_server_size = {
+                "instanceSizeType": "custom",
+                "instanceSize": {"milliCpus": 1000, "memoryMb": 4096},
+            }
+        payload["defaultServerSize"] = default_server_size
```
acroz, line 327: There might be some common schemas with the `ServerClient` might be worth considering reusing?

faculty/clients/api.py
```diff
@@ -0,0 +1,635 @@
...
+        self,
+        project_id,
+        api_id,
+        environment_ids=None,
+        api_definition=None,
+        server_type=None,
+        current_api=None,
+    ):
+        if not current_api:
+            current_api = self._get_current_api_definition(project_id, api_id)
```
acroz, line 378: This kind of logic should be outside the client - the client methods are intended to map exactly on to single endpoints.

faculty/clients/api.py
```diff
@@ -0,0 +1,635 @@
...
+
+        # TODO: turn these into schema handling as well
+        if default_server_size is None:
+            default_server_size = {
+                "instanceSizeType": "custom",
+                "instanceSize": {"milliCpus": 1000, "memoryMb": 4096},
+            }
+        payload["defaultServerSize"] = default_server_size
+
+        return self._post_raw(
```
acroz, line 329: No need (and unconventional in this library) to return the Flask Response object:
```suggestion
        self._post_raw(
```

faculty/clients/api.py
```diff
@@ -0,0 +1,635 @@
...
+        if environment_ids:
+            payload["environmentIds"] = environment_ids
+
+        if api_definition:
+            payload["definition"] = api_definition
+        endpoint = "/project/{}/api/{}/definition".format(project_id, api_id)
+        return self._put(endpoint, _APISchema(), json=payload)
+
+    def delete(self, project_id, api_id):
+        return self._delete_raw(
```
acroz, line 394: As above:
```suggestion
       self._delete_raw(
```

faculty/clients/api.py
```diff
@@ -0,0 +1,635 @@
...
+        label : str, optional
+            If provided, set this value as the new key's label.
+
+        Returns
+        -------
+        APIKey
+            The newly created key.
+        """
+        endpoint = "/project/{}/api/{}/key".format(project_id, api_id)
+        # TODO: maybe more sensible default, the UI doesn't allow empty name but the api accepts this
```
acroz, line 434: Does the API accept a missing (as opposed to empty) label? If so, I'd suggest not setting a default here, if not, I'd suggest forcing the caller to provide one.

faculty/clients/api.py
```diff
@@ -0,0 +1,635 @@
...
+        self,
+        project_id,
+        api_id,
+        key_id,
+        server_type=None,
+        image_version=None,
+        current_api=None,
+    ):
+        if not current_api:
+            current_api = self._get_current_api_definition(project_id, api_id)
```
acroz, line 567: As above

faculty/clients/api.py
```diff
@@ -0,0 +1,635 @@
...
+            script_name=script_name,
+        )
+
+    @staticmethod
+    def script_definition(working_directory, script_name):
+        return ScriptDefinition(
+            api_type=APIType.SCRIPT,
+            working_directory=working_directory,
+            script_name=script_name,
+        )
```
acroz, line 356: Is this point of these to aid construction of API definition objects to send to the backend? I don't think these belong as static methods of the client.

faculty/clients/api.py
```diff
@@ -0,0 +1,635 @@
...
+from collections import namedtuple
+from enum import Enum
+
+from marshmallow import fields, post_load
+from marshmallow_enum import EnumField
+
+from faculty.clients.base import BaseClient, BaseSchema
+from faculty.clients.server import _ServerSchema as ServerSchema
+
+Instance = namedtuple("Instance", ["instance", "outdated", "key"])
```
acroz, line 27: Newer code should use `attrs` rather than `namedtuple` for new data classes. See, for example, the job client - note in particular that the longer form imports (`from attr import attrs, attrib`) are preferred.

faculty/clients/api.py
```diff
@@ -0,0 +1,635 @@
...
+    script_name = fields.String(data_key="scriptName")
+
+    @post_load
+    def make_command_definition(self, data, **kwargs):
+        if data["api_type"] == "wsgi":
+            return WSGIDefinition(**data)
+        elif data["api_type"] == "script":
+            return ScriptDefinition(**data)
+        elif data["api_type"] == "plumber":
+            return PlumberDefinition(**data)
```
acroz, line 194: It's a bit odd that we're setting an api_type field on these definition objects despite the fact that this information is already encoded in the type. I think what we really want is to use something like `marshmallow-oneofschema` to serialise/deserialised between the correct types - note that we have a version of `marshmallow-oneofschema` vendored in this library.

faculty/clients/api.py
imrehg: Yeah, there are some stuff here and there, there's `_InstanceSizeSchema` from `JobClient` for example, but will see to mix and match further. It would be great to avoid duplication, there are some subtle differences as much as I can tell, but still digging into the actual platform api schema.

faculty/clients/api.py
imrehg: @acroz while I was checking this out, came across the more modern use of `attr` with `define` and `field` rather than `attrs` and `attrib` (that are the "classic" versions and actually used underneath). https://www.attrs.org/en/stable/api.html#api-reference

Would it make sense to use the new define/field way for newer code (and possibly migrate the older code to that form as well? It feels like going with the classic form of attrs/attrib just for consistency sake in other modules is inconsistent how the SDK moved to using this in place of namedtuples in the first place (ie. sensible change is good).

Anyways I'm still looking at the differences a bit, the new terminology seems more self-explanatory so far...

faculty/clients/api.py
imrehg: Missing value is not accepted, will go with the second option of making it a mandatory input.

faculty/clients/api.py
imrehg: I just side-stepped this by not loading the API, though it might be a little less ergonomic even if simpler internally.

faculty/clients/api.py
imrehg: Removed all of these, people can create the `WSGIDefinition`/`PlumberDefinition`/`ScriptDefinition` objects themselves to pass onto the relevant functions.

faculty/clients/api.py
imrehg: Redone it with the oneofschema pattern, it's pretty neat :)

faculty/clients/api.py
```diff
@@ -0,0 +1,745 @@
...
+from faculty._oneofschema import OneOfSchema
+from faculty.clients.base import BaseClient, BaseSchema
+from faculty.clients.job import InstanceSize, _InstanceSizeSchema
+from faculty.clients.server import Server, _ServerSchema
+
+
+class DeploymentStatus(Enum):
+    """An enumeration of possible API deployment statuses."""
+
+    NOTDEPLOYED = "not-deployed"
```
acroz, line 36:
```suggestion
    NOT_DEPLOYED = "not-deployed"
```

faculty/clients/api.py
```diff
@@ -0,0 +1,745 @@
...
+    STARTING = "starting"
+    DEPLOYED = "deployed"
+    ERROR = "error"
+
+
+@define
```
acroz, line 42: Rather than attrs new interface, I'd actually advocate for using dataclasses now that we're only supporting 3.6+ (dataclasses are 3.7+, but there is a 3.6 backport that [you can install on Python 3.6 only](https://github.com/acroz/pylivy/blob/master/setup.py#L28)). The syntax is basically the same but this has the obvious advantage of being in the standard library and better integrated with other Python tools.

I'd suggest we use dataclasses here and gradually migrate other clients to use dataclasses in future. What do you think?

faculty/clients/api.py
```diff
@@ -0,0 +1,745 @@
...
+    instance_size_type: str
+    instance_size: Optional[InstanceSize] = None
+
+
+@define
+class WSGIDefinition:
+    working_directory: str
+    module: str
+    wsgi_object: str
+    last_updated_at: Optional[datetime] = None
```
acroz, line 66: From looking at the aperture code, I don't think this is `Optional` (see ControllerReads line 92 and PersistentDetails line 43).

faculty/clients/api.py
```diff
@@ -0,0 +1,745 @@
...
+
+
+@define
+class PlumberDefinition:
+    working_directory: str
+    script_name: str
+    last_updated_at: Optional[datetime] = None
```
acroz, line 73: Ditto

faculty/clients/api.py
```diff
@@ -0,0 +1,745 @@
...
+
+
+@define
+class ScriptDefinition:
+    working_directory: str
+    script_name: str
+    last_updated_at: Optional[datetime] = None
```
acroz, line 80: Ditto

acroz (changes requested):

faculty/clients/api.py
imrehg: Yeah, I think this is a good call. Don't have strong reason to go with `attr`, not given how do we use things.
The plan sounds very sensible, will do the changes here, and once merged, will create an issue to track the whole migration as you outline.

faculty/clients/api.py
imrehg: It is optional, because this class is used not just for replies from `aperture`, but also request for creating APIs, and then the user would need to supply a `last_updated_at` value.

ie. with the current `Optional` setting, we can do

```python
api_definition = WSGIDefinition(working_directory="/project/testapi", module="app", wsgi_object="app")
# [...snip...]
x = client.create(project_id=PROJECT_ID, command_definition=api_definition, subdomain=subdomain, name=name, default_server_size=server)
```
and without it the request will fail, obviously, unless a value is set, even if the `create` endpoint doesn't require that value (see the Command Definitions in the README).

Thus the tricky place is that for requests it's definitely optional, for replies it will always be included, and thus the reason why I did it like this.

Given all the above, how would you prefer to resolve it?

* keeping it `Optional` due to the dual use?
* making it non-optional but having a default value? (of `datetime.now()` or something?) (this feels dirty)
* making it non-optional and make people pass in some value? (feels dirty like above)
* something completely different?

acroz: Hmm, the "right" thing to do here would be to have different models, since the schema for the create body is different than the schema for the get response. But that will cause a fair bit of duplication, which is obviously undesirable. I'll have a think and get back to you.

I'd say a definite no to the second and third bullets in your comment.

imrehg: Hey @acroz, I've pushed some changes to see if they resolve this conundrum, breaking classes up into "Something" (for requests) and "SomethingResponse" (with inheritance derived from Something, for responses). This _feels_ like having the right amount of duplication (ie. "SomethingResponse" only needs the additional or changed fields). I've tested it very briefly so far and seems to work, also the way we are creating the `json` payload, that's helpful that we don't need to convert Something <-> SomethingResponse at any point, and still can use them. (that conversion would be a bit verbose and hacky, as much as I've researched).

It's not fully fledged yet (ie. there might be mistakes in the rewrite), but wanted to get your opinion with code.

PR #200, opened 2022-05-04
update "black" dependency as it was breaking on its own dependencies

imrehg:
* black: this issue: https://github.com/psf/black/issues/2634
* expanding Python version test coverage
* adjusting `setuptools_scm` use as per recommended by their docs: https://github.com/pypa/setuptools_scm#pyprojecttoml-usage

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,3 @@ line-length = 79
 
 [tool.setuptools_scm]
 version_scheme = "post-release"
-
-[build-system]
-requires = ["setuptools>=45", "wheel", "setuptools_scm>=6.2"]
-build-backend = "setuptools.build_meta"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,9 @@
 [tool.black]
 line-length = 79
+
+[tool.setuptools_scm]
+version_scheme = "post-release"
+
+[build-system]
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -1,2 +1,9 @@
 [tool.black]
 line-length = 79
+
+[tool.setuptools_scm]
+version_scheme = "post-release"
+
+[build-system]
```
liamcoatman, line 7: Do we need this? I thought it was only needed if `pyproject.toml` was being used by `pip` to build the package rather than `setup.py`?

acroz:
LGTM. I've happy to merge once Liam's query is resolved (I am personally unopinionated on whether we include the build-system config or not - we are currently just using setup.py to build as he said).

pyproject.toml
imrehg: @liamcoatman it is a good challenge. I re-read their recommendations' meanings, and tried install without this part. I'm _pretty confident_ now that indeed it's not needed in the current setup, so pushed an extra change with this removed.

imrehg:
@acroz resolved for now (I have slight favor for keeping it in, but not enough support for needing it, so removed for now). Thank you!

acroz (approved):

Issue #202, opened 2022-05-14
Consider dropping marshmallow_enum from dependencies


imrehg:
The project seems to be abandoned (no change since 2019), and has potential to break things in the future, with PRs like this https://github.com/justanr/marshmallow_enum/pull/49 pending.

Something along the lines of https://github.com/marshmallow-code/marshmallow/issues/1470#issuecomment-563114297 might be applicable, though haven't tried to make it work on this codebase much yet.

PR #203, opened 2022-05-14
marshmallow schema maintenance to address RemovedInMarshmallow4Warning results

imrehg:
Since the dependency list do not limit the `marshmallow` version, it is likely prudent to fix the multitude of `RemovedInMarshmallow4Warning` results that the tests are throwing. Most of them are pretty straightforward, but happy to elaborate.

Tests were run with `tox -e py38 -- -Werror` to make warnings into errors for the process.

There's one remaining warning that is coming from the `marshmallow_enum` library, and #202 should be relevant for this.

### DIFF
--- a/faculty/clients/base.py
+++ b/faculty/clients/base.py
@@ -363,8 +363,8 @@ class BaseSchema(Schema):
 
 
 class _ErrorSchema(BaseSchema):
-    error = fields.String(missing=None)
-    error_code = fields.String(data_key="errorCode", missing=None)
+    error = fields.String(load_default=None)
+    error_code = fields.String(data_key="errorCode", load_default=None)
 
 
 def _check_status(response):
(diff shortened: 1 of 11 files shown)


acroz:
Hey, thanks for this update. Looking at [the changelog](https://marshmallow.readthedocs.io/en/stable/changelog.html), it looks like `load_default` was introduced in 3.13.0. If we merge this PR, we should specify this in our dependencies. Are we ok with requiring 3.13.0 or later? (FWIW I'm not saying we can't - but might be worthwhile for you to think through).

For the sake of completeness, your other updates depend on earlier updates to the API (make_error: 3.0.0, self -> lambda: 3.3.0) so an updated min version to 3.13.0 should cover all.

imrehg:
@acroz that's a good shout, didn't think of that aspect! Given that the release is almost 1 year old, it should be on the safe territory, but I'll check for any side-effects or other considerations, and add the minimum version if it looks okay. Will reping when that's done.

